PR #8, opened 2021-08-04
Make sure that subprocess uses the same python interpreter

timoschick:
The `test_train` failed for me because `subprocess.Popen("python", ...)` does not ensure that the same virtual environment is used that was used to run the test. This PR fixes that issue.


SaulLu (approved):
Sorry, you're absolutely right it's better if it runs with the right dependencies!

Sounds perfect to me!

PR #9, opened 2021-08-05
Create a `get_dataloader` for dataset with metadata

### LINKED ISSUE #2: Complete `get_dataloader` method
# Initial thoughts

I think that the `get_dataloader` method is in charge of:
- retrieving the dataset contained in one (or more?) `.jsonl` file with the template below
- ~applying collation functions and tokenize the text (in what order?)~ EDIT: for now, augment the text with metadata and tokenizer it as a pre-processing step
- returning train and evaluation dataloaders 

Toy `.jsonl` data file
```JSONL
[
    {
        "document_id": 10,
        "text": "this is the input",
        "metadata": [
            {
                "key": "url",
                "type": "global",
                "value": "http://1"
            },
            {
                "key": "entity",
                "type": "local",
                "value": "address",
                "start_idx": 20,
                "end_idx": 40
            }
        ]
    },
    {
        "document_id": 12,
        "text": "this is the second input",
        "metadata": [
            {
                "key": "url",
                "type": "global",
                "value": "http://2"
            },
            {
                "key": "entity",
                "type": "local",
                "value": "date",
                "start_idx": 60,
                "end_idx": 90
            }
        ]
    }
]
```

SaulLu:
As discussed, this PR aims to plug the two existing parts of the code together.

More concretely, this PR introduces a new function to retrieve dataloaders containing examples augmented with metadata. As discussed, the approach taken here is to perform this processing as a pre-processing step. If necessary, we can modify the approach later to perform this processing on-the-fly.

This PR also contains tests for this new method and the whole training.

I want to draw your attention to a **structural change** introduced in this PR and that we need to discuss. When I wanted to add new tests for my new `get_dataloaders` function, I started to run into a lot of problems with the relative paths between files. The solution that seemed easiest to me was to create a **package** that we can then install in editable mode in our virtual environment and test. Are you happy with this change?

Again, I'd love to have your feedbacks and thoughts!

Fix #2


metadata/experiments/with_metadata.py
```diff
@@ -0,0 +1,135 @@
...
+            )
+    else:
+        data_files = {}
+        if args.train_file is not None:
+            data_files["train"] = args.train_file
+        if args.validation_file is not None:
+            data_files["validation"] = args.validation_file
+        extension = args.train_file.split(".")[-1]
+        if extension == "txt":
+            extension = "text"
```
timoschick, line 73: Do we really want to support text-only files for experiments with metadata?

setup.py
```diff
@@ -0,0 +1,22 @@
...
+setup(
+    name="MyPackageName",
+    version="0.0.0",
+    url="https://github.com/bigscience-workshop/metadata.git",
+    author="Author Name",
+    author_email="author@gmail.com",
+    description="Description of my package",
+    packages=find_packages(),
+    install_requires=install_requires,
+)
```
timoschick, line 22: Let's discuss the values to use here in our meeting today (We may follow the conventions used by other bigscience repositories, e.g., [promptsource](https://github.com/bigscience-workshop/promptsource/blob/main/setup.py) and [bigscience](https://github.com/bigscience-workshop/bigscience/blob/master/setup.py)).

timoschick (approved):
LGTM. Creating a package absolutely makes sense, thanks for taking care of this! 👍

metadata/experiments/with_metadata.py
SaulLu: You are right, the script to load from text will not work! I'm doing a new commit to return an error in this case

setup.py
SaulLu: Great idea! :)

cccntu (approved):
LGTM

PR #10, opened 2021-08-05
Modify values for package setup

timoschick:
This PR modifies package-level metadata. Following https://stackoverflow.com/questions/2852283/python-naming-a-module-that-has-a-two-word-name , I decided to go with "bsmetadata" (short for BigScience Metadata) as a package name, but I'm open to any suggestions for alternative names.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,14 +9,13 @@ def req_file(filename):
 
 install_requires = req_file("requirements.txt")
 
-# TODO Fill
 setup(
-    name="MyPackageName",
+    name="bsmetadata",
     version="0.0.0",
     url="https://github.com/bigscience-workshop/metadata.git",
-    author="Author Name",
-    author_email="author@gmail.com",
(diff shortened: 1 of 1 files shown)


cccntu (approved):
LGTM

SaulLu (approved):
Thanks a lot for asking about the package names, it sounds great!

By the way, I wonder if we should rename in the code everything that was called `metadata` by `bsmetadata` ?

timoschick:
> By the way, I wonder if we should rename in the code everything that was called `metadata` by `bsmetadata` ?

Good point, I think we should. Do you want to take care of this or should I do that?

SaulLu:
> > By the way, I wonder if we should rename in the code everything that was called `metadata` by `bsmetadata` ?
> 
> Good point, I think we should. Do you want to take care of this or should I do that?

I can handle it if you want, I open an issue :)

PR #14, opened 2021-08-05
Add help message

cccntu:
Use `h=True` to print help message.
```sh
python metadata/train.py h=True
```
I checked `transformers/src/transformers/hf_argparser.py` to understand how to access the metadata.
Need to add support for nested dataclasses.

I followed suggestion 4 in here:
https://github.com/facebookresearch/hydra/issues/633

Edit:
I think we should do this, so users can use `--help`, `-h`
```diff
if __name__ == "__main__":
+  if '--help' in sys.argv or '-h' in sys.argv:
+    show_help()
+    sys.exit()
  main()
```


timoschick:
A quick fix to support nested dataclasses would be to recursively check `dataclasses.is_dataclass(field)` for each `field` (and if so, call the print logic with that `field` as config). I don't see any better way if we want to stick with hydra, but I'd be ok with this approach given that it takes much less time than rewriting everthing to work without hydra. What do you think, @SaulLu @cccntu ?

timoschick:
Ah, I see you've already done just that, @cccntu. Great, thanks! 👍

cccntu:
I just did that! 😄 @timoschick 
I think the help message still needs a header, also all the comments need to be changed to use `field`. 
Otherwise I think this can be merged. Do you want to take over?

timoschick:
Sure, I can do that!

SaulLu:
Thank you very much for looking into all the possibilities. The one you chose seems to be the best for our case. :+1:

PR #16, opened 2021-08-06
Randomly include metadata with a given probability

timoschick:
This PR adds a parameter `metadata_probability` to the `DataConfig`, which controls the probability of adding metadata to an input example.
Right now, whether to add metadata is decided once for each input (as opposed to making individual decisions for each chunk) - we might want to change this later.


SaulLu (approved):
Thank you for this addition! Everything looks good to me :+1:

cccntu (approved):
LGTM, I think per-example is good enough, arguably better. 👍

PR #18, opened 2021-08-17
feat: add `UrlProcessor` for applying `unquote_plus()` to URLs.

tianjianjiang:
I will use code review comments to describe what I've done with this PR.

### DIFF
--- a/bsmetadata/metadata_processors.py
+++ b/bsmetadata/metadata_processors.py
@@ -15,6 +15,7 @@ This script provides functions for processing different kinds of metadata.
 """
 import datetime
 from typing import Any, Dict, Optional, Tuple
+from urllib.parse import unquote_plus
 
 from bsmetadata.input_pipeline import DataConfig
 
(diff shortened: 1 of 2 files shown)


bsmetadata/metadata_processors.py
```diff
@@ -38,11 +39,13 @@ def process_global(self, metadata_attrs: Dict[str, Any]) -> Optional[str]:
         prefix will be "url: wikipedia.com/Apple".
 
         Args:
-            metadata_attrs: All attributes of this metadata instance. Each global metadata instance is expected to have an attribute "key"
-            of type string and a corresponding "value" of arbitrary type.
+            metadata_attrs (:obj:`Dict[str, Any]): All attributes of this metadata instance.
```
tianjianjiang, line 42: Docstring changes like this are just adding types like other HuggingFace repos do.

tests/test_metadata_utils.py
```diff
@@ -50,6 +50,13 @@ def setUp(self) -> None:
                     {"key": "html", "type": "local", "value": "i", "char_start_idx": 43, "char_end_idx": 48},
                 ],
             },
+            {
+                "id": "0003",
+                "text": "Wubba Lubba Dub Dub!",
+                "metadata": [
+                    {"key": "url", "type": "global", "value": "callto:RickAndMorty/Year%202021/"},
```
tianjianjiang, line 57: `callto:` isn't really necessary, just want to ensure the implementation isn't vulnerable to rare URI schemes.

bsmetadata/metadata_processors.py
timoschick: I'm not an expert on type hinting in python, but shouldn't it be sufficient that we already have the types in the method signature (e.g., `def process_global(self, metadata_attrs: Dict[str, Any]) -> Optional[str]:`)? I would think that storing this information in the signature **and** the docstring makes things more error prone as it's likely that someone changing the method signature will forget to adapt it in one of the two places.

timoschick: Oh and just a nitpick, there's a single `` ` `` missing after `` `Dict[str, Any]``

tests/test_metadata_utils.py
timoschick: Sure, makes sense to check for such corner cases 👍

bsmetadata/metadata_processors.py
tianjianjiang: >  storing this information in the signature and the docstring makes things more error prone

Indeed. I wasn't really considering the side effect but merely applying the practice of HuggingFace repos.


> there's a single ` missing after `Dict[str, Any]

A case in point 😅

(I'm gonna alter the history to remove docstring changes.)

PR #19, opened 2021-08-17
build: set Python and dependency versions with an optional Poetry env

tianjianjiang:
I will use code review comments to describe what I've done for the PR.

~~Three~~Two side notes:
1. Python 3.7.11 is for Colab;~~Although python>=3.8.10 also supports Apple Silicon and BigSur, 3.9 syntax is nicer for Black formatting;~~
~~2. GitHub Action doesn't have a roadmap for Apple M1 VM, therefore `test.yml` isn't really for the case;~~
1. Poetry is optional for managing venv and dependencies, but syncing with requirements(-dev).txt must be done manually for the time being.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Code quality
+name: Test
 
 on:
     push:
@@ -11,8 +11,8 @@ on:
         types:
             - created
 jobs:
-    check_code_quality:
-        name: Check code quality
+    test:
+        name: Test
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v2
@@ -23,7 +23,9 @@ jobs:
             - name: Install dependencies
               run: |
                   python -m pip install --upgrade pip
+                  python -m pip install .
                   python -m pip install -r requirements-dev.txt
-            - name: Make quality
+            - name: Test
               run: |
-                  make quality
+                  python -m pytest tests/test_get_dataloaders.py
+                  python -m pytest tests/test_metadata_utils.py
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,28 @@
 [tool.black]
 line-length = 119
-target-version = ['py35']
+target-version = ['py37']
+
+[tool.poetry]
+name = "metadata"
+version = "0.1.0"
+description = "Codebase for including metadata (e.g., URLs, timestamps, HTML tags) during language model pretraining."
+authors = ["Your Name <you@example.com>"]
+
+[tool.poetry.dependencies]
+python = ">=3.7.11, <3.10"
+torch = "1.8.1"  # 1.9 also works for Colab TPU, but not for GPU/CPU; the bugfix of 1.8.2 seems irrelevant for now.
+hydra-core = ">=1.1,<1.2"  # Poetry will likely update it to 1.1.1, but it is probably ok and good for bugfixes.
+wandb = ">=0.10.32,<1"  # Poetry will likely update it to 0.12.1, but it is probably ok and good for bugfixes.
+transformers = "^4.6.0"  # Poetry will likely update it to 4.10.0, but it is probably ok and good for bugfixes.
+accelerate = ">=0.4.0,<1"  # We may want to use 0.5.0 in the near future
+datasets = {extras = ["streaming"], version = "^1.11.0"}
+
+[tool.poetry.dev-dependencies]
+isort = "^5.5.4"
+black = "^21.4b0"
+flake8 = "^3.8.3"
+pytest = "^6.2.4"
+
+[build-system]
+requires = ["poetry-core>=1.0.0", "setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-torch==1.8.1
-hydra_core==1.0.6
-wandb==0.10.30
-transformers==4.6.0
-accelerate==0.3.0
-datasets==1.11.0
+torch==1.8.1  # 1.9 also works for Colab TPU, but not for GPU/CPU; the bugfix of 1.8.2 seems irrelevant for now.
+hydra_core>=1.1,<1.2  # pip will likely update it to 1.1.1, but it is probably ok and good for bugfixes.
+wandb>=0.10.32,<1  # pip will likely update it to 0.12.1, but it is probably ok and good for bugfixes.
+transformers>=4.6.0,<5  # pip will likely update it to 4.10.0, but it is probably ok and good for bugfixes.
+accelerate>=0.4.0,<1  # We may want to use 0.5.0 in the near future
+datasets[streaming]>=1.11.0,<2
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,8 @@ install_requires = req_file("requirements.txt")
 
 setup(
     name="bsmetadata",
-    version="0.0.0",
+    python_requires=">=3.7.11, <3.10",
+    version="0.1.0",
     url="https://github.com/bigscience-workshop/metadata.git",
     author="Multiple Authors",
     author_email="xxx",
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,0 +1,1680 @@
+[[package]]
+name = "accelerate"
+version = "0.4.0"
+description = "Accelerate"
+category = "main"
+optional = false
+python-versions = ">=3.6.0"
+
+[package.dependencies]
+pyyaml = "*"
+torch = ">=1.4.0"
+
+[package.extras]
+docs = ["docutils (==0.16.0)", "recommonmark", "sphinx (==3.2.1)", "sphinx-markdown-tables", "sphinx-rtd-theme (==0.4.3)", "sphinx-copybutton", "sphinxext-opengraph (==0.4.1)"]
+quality = ["black (==21.4b0)", "isort (>=5.5.4)", "flake8 (>=3.8.3)"]
+sagemaker = ["sagemaker"]
+test = ["pytest", "pytest-xdist"]
+
+[[package]]
+name = "aiohttp"
+version = "3.7.4.post0"
+description = "Async http client/server framework (asyncio)"
+category = "main"
+optional = false
+python-versions = ">=3.6"
+
+[package.dependencies]
+async-timeout = ">=3.0,<4.0"
+attrs = ">=17.3.0"
+chardet = ">=2.0,<5.0"
+multidict = ">=4.5,<7.0"
+typing-extensions = ">=3.6.5"
+yarl = ">=1.0,<2.0"
+
+[package.extras]
+speedups = ["aiodns", "brotlipy", "cchardet"]
+
+[[package]]
+name = "antlr4-python3-runtime"
+version = "4.8"
+description = "ANTLR 4.8 runtime for Python 3.7"
+category = "main"
+optional = false
+python-versions = "*"
+
+[[package]]
+name = "async-timeout"
+version = "3.0.1"
+description = "Timeout context manager for asyncio programs"
+category = "main"
+optional = false
+python-versions = ">=3.5.3"
+
+[[package]]
+name = "atomicwrites"
+version = "1.4.0"
+description = "Atomic file writes."
+category = "dev"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+
+[[package]]
+name = "attrs"
+version = "21.2.0"
+description = "Classes Without Boilerplate"
+category = "main"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+
+[package.extras]
+dev = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "furo", "sphinx", "sphinx-notfound-page", "pre-commit"]
+docs = ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"]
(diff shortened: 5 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
...
+                  python-version: 3.9
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip
+                  python -m pip install .
+                  python -m pip install -r requirements-dev.txt
+            - name: Test
+              run: |
+                  python -m pytest tests/test_get_dataloaders.py
+                  python -m pytest tests/test_metadata_utils.py
```
tianjianjiang, line 31: `test_train.py` took about 5 minutes to complete so excluded for now.

pyproject.toml
```diff
@@ -1,3 +1,29 @@
...
+target-version = ['py39']
+
+[tool.poetry]
+name = "metadata"
+version = "0.1.0"
+description = "Codebase for including metadata (e.g., URLs, timestamps, HTML tags) during language model pretraining."
+authors = ["Your Name <you@example.com>"]
+
+[tool.poetry.dependencies]
+python = "~3.9"
```
tianjianjiang, line 12: Enforcing `3.9` if using Poetry.

setup.py
```diff
@@ -11,7 +11,8 @@ def req_file(filename):
 
 setup(
     name="bsmetadata",
-    version="0.0.0",
+    python_requires=">=3.9.1",
```
tianjianjiang, line 14: Enforcing `>=3.9.1` if using setuptools.

pyproject.toml
```diff
@@ -1,3 +1,29 @@
...
+target-version = ['py39']
+
+[tool.poetry]
+name = "metadata"
+version = "0.1.0"
+description = "Codebase for including metadata (e.g., URLs, timestamps, HTML tags) during language model pretraining."
+authors = ["Your Name <you@example.com>"]
+
+[tool.poetry.dependencies]
+python = ">=3.9.1, <3.10"
```
tianjianjiang, line 12: Enforcing Apple Silicon compatible py39 if using Poetry.

timoschick:
@SaulLu Since you've been setting up our current workflows and are more experienced with this than I am, could you perform the review for this pull request?

pyproject.toml
```diff
@@ -1,3 +1,29 @@
...
 line-length = 119
-target-version = ['py35']
\ No newline at end of file
+target-version = ['py39']
+
+[tool.poetry]
+name = "metadata"
+version = "0.1.0"
+description = "Codebase for including metadata (e.g., URLs, timestamps, HTML tags) during language model pretraining."
+authors = ["Your Name <you@example.com>"]
```
SaulLu, line 9: Just a nitpick, I have the impression that it is not the same thing as what is put in `setup.py`. Is it voluntary? :slightly_smiling_face:

SaulLu (approved):
Thank you very much for working on these additions. :+1: 

I had heard good things about poetry but never used it personally until now. If you don't mind, I have two small questions about poetry: 
- Can you confirm that the information is currently doubled between `poetry.lock` and (`requirements.txt`, `requirements-dev.txt` and èsetup.py`)? If this is the case, it's not a problem for me at all but I think we might benefit from stated it somewhere (in comments or on the readme).
- If I'm not mistaken, I understand that poetry fixes the virtual env packages (in the `poetry.lock` file). Knowing that torch requires special installation depending on the accelerator used, do you know how to handle this with poetry? 

Finally, I just want to say that if we run experiments on Jean Zay, we may have to change these requirements (python 3.9, and the versions of external libraries).

.github/workflows/test.yml
SaulLu: Thank you very much for this addition!

tianjianjiang: @SaulLu You're very welcome! I hope this will help a little without too much intrusion...

tianjianjiang:
> Thank you very much for working on these additions. 👍

My pleasure 👍

----
> * Can you confirm that the information is currently doubled between `poetry.lock` and (`requirements.txt`, `requirements-dev.txt` and èsetup.py`)?

Partially yes. Although poetry.lock covers requirements.txt, requirements-dev.txt, and their dependencies, Poetry doesn't work closely with setup.py and setuptools. Some more info about it below but feel free to ignore it since it isn't really a big issue for experiments.

<details>
  <summary>[Poetry, pip-tools, and setuptools]</summary>
  I was a fan of pip-tools because not only it calculates deps fast and accurately, but also flexible for separating dev requirements. Everything is controllable except everything is manual. 😛
  Poetry is based on pip-tools, but the price of automation is that it isn't as flexible as pip-tools is. In addition, it doesn't want to process setup.py, so there is an edge case when certain deps is defined in setup.py, Poetry won't resolve it. (It happened to me once: python-poetry/poetry#1287)
  Therefore, I sometimes keep both setup.[py|cfg] and poetry.lock, and then I will also have to remember to sync requirements.txt (or constraints.txt) manually, and then hope one day somebody will improve the situation...
</details>

----
> If this is the case, it's not a problem for me at all but I think we might benefit from stated it somewhere (in comments or on the readme).

I will update README as soon as possible. In short, Poetry is likely something auxiliary.

----
> * If I'm not mistaken, I understand that poetry fixes the virtual env packages (in the `poetry.lock` file). Knowing that torch requires special installation depending on the accelerator used, do you know how to handle this with poetry?

Poetry reads pyproject.toml and then generates poetry.lock. As long as torch and accelerator versions can be defined by requirements.txt, same thing can happen to pyproject.toml, and then Poetry can handle them (except for the edge case aforementioned in [Poetry, pip-tools, and setuptools], python-poetry/poetry#1287).

----
> Finally, I just want to say that if we run experiments on Jean Zay, we may have to change these requirements (python 3.9, and the versions of external libraries).

I also expect that to happen in the near future. For this PR, I guess we can at least settle with Python version. If none of us need Apple M1 and Jean Zay requires older Python, I can downgrade it right away.

pyproject.toml
SaulLu: Indeed, I was talking about the `author` / `authors` section.

Thanks for the link, I now understand that the section keys differ between `poetry` and `Setuptools`

SaulLu:
Thank you very much for all these explanations! :+1: 

> I also expect that to happen in the near future. For this PR, I guess we can at least settle with Python version. If none of us need Apple M1 and Jean Zay requires older Python, I can downgrade it right away.

At the moment, it's a bit early to tell about JZ. We can leave the python 3.9 version and come back down if necessary later?

tianjianjiang:
@SaulLu

> At the moment, it's a bit early to tell about JZ. We can leave the python 3.9 version and come back down if necessary later?
> 
> EDIT: I have the impression that the [WG that runs experiments for the big language model uses python=3.8](https://github.com/bigscience-workshop/bigscience/tree/master/jz/envs#creating-production-conda-env), It might be easier to go with this version, what do you think?

I thought so (for py38) but then intriguingly modeling-promptsource did [python 3.8 -> >=3.7](https://github.com/bigscience-workshop/promptsource/commit/9b91d27d50d0f82bce6922156240a1e3a1510ba0) for a reason that I don't know (I haven't asked). I'm gonna ask around and see if we can have a more definite answer.

tianjianjiang:
@timoschick @SaulLu In case package dependencies also concern you with data tooling, some additional info before your meeting:
- I guess many of us want to use Colab, and there are some minor dependency conflicts
    ```
    torchvision 0.10.0+cu102 has requirement torch==1.9.0, but you have torch 1.8.1.
    torchtext 0.10.0 has requirement torch==1.9.0, but you have torch 1.8.1.
    ```
- Besides `torch==1.9.0+cu102`, Colab uses Python 3.7.11.
- Since `accelerate==1.4.0`, it begins to support DeepSpeed that I believe some WG(s) may be working on it.

I can fix the above situations easily, as long as no serious conflicts with any of our local or Jean Zay envs.

tianjianjiang:
Another new (actually old but I just forgot about it) info: https://github.com/bigscience-workshop/evaluation/pull/58#issuecomment-903781001

tianjianjiang:
Colab (CPU) snippet and output:
```bash
!git clone -b build-use_py39_for_apple_silicon_with_optional_poetry https://github.com/tianjianjiang/bigscience-metadata.git
%cd bigscience-metadata
!pip install -q -e .
!TOKENIZERS_PARALLELISM=false python -m pytest .
```

```
============================= test session starts ==============================
platform linux -- Python 3.7.11, pytest-3.6.4, py-1.10.0, pluggy-0.7.1
rootdir: /content/bigscience-metadata, inifile:
plugins: typeguard-2.7.1, hydra-core-1.0.6
collected 14 items                                                             

tests/test_get_dataloaders.py .......                                    [ 50%]
tests/test_metadata_utils.py .....                                       [ 85%]
tests/test_train.py ..                                                   [100%]

========================= 14 passed in 327.71 seconds ==========================
```

tianjianjiang:
@SaulLu 
I'm gonna merge this soon if no further concerns.

SaulLu:
I apologise for answering so late, I didn't answer at the time because I don't really have a solution and unforthunatly I still don't. 

I understand all your points and indeed it seems to me a very good idea to be compatible with google collaboratory.

I can try to work on JZ with the dependencies that arrange on google collaboratory (for the moment I use `python 3.8` and `torch 1.8.1`). I will re-send a PR if there is any incompatibility for any reason :) 

I'm taking the opportunity to make 2 discuss the need to change the version of 2 other packages:
- for H, I think we would need the V version to be able to use `OmegaConf.to_object` which would allow transforming `args` which was a `omegaconf.dictconfig.DictConfig` into another object whose hash is deterministic. This is usefull for `datasets` library caching
- `wandb` I would like to use at least the V version in which they have released a feature that allows to synchronize offline runs that are still in progress

tianjianjiang:
Hi @SaulLu ,
> I apologise for answering so late, I didn't answer at the time because I don't really have a solution (to combine compatibility with google collaboratory and use the same dependencies as WG engineering on JZ) and unforthunatly I still don't.

No problems at all. I was also unable to reply in almost 2 weeks.
I did actually have a plan B but hasn't played with it for quite a while, so not really confident if it will still work.
The plan B is that we may be able to use ngrok to connect to Colab. There
Honestly I don't really have a strong preference towards 3.7 (or 3.8). The only reason is that I'm lazy about trying out the plan B, or plan C, D, or E, etc., such as replacing the whole python environment of Colab, applying TPU Research Cloud (TRC) program, or convincing my boss to use paid resources.

> I can try to work on JZ with the dependencies that arrange on google collaboratory (for the moment I use `python 3.8` and `torch 1.8.1`). I will re-send a PR if there is any incompatibility for any reason :)

I think `torch` here is probably a more critical dependency than python itself. I believe that python can still be `>=3.7` for our package so both JZ and Colab will be fine, but `torch` may have to be specific. I guess `torch 1.8.1` will be just fine, even though `accelerate`'s examples are using `torch 1.9` now. Let me make a quick test on Colab and get back to you ASAP.

> * `hydra_core`, I think we would need the `1.1.0` ...
> * `wandb` I would like to use at least the `0.10.32` ...

I will also try them out on Colab together with `torch`, `torch-xla`, and `cloud-tpu-client`.

tianjianjiang:
So far the behavior of Colab TPU seems similar to my previous setup, except for a harmless warning:
```
bsmetadata/train.py:107: UserWarning: 
config_path is not specified in @hydra.main().
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/changes_to_hydra_main_config_path for more information.
  @hydra.main(config_name="config")
```

SaulLu:
Thank you very much for checking it out! :raised_hands: 

> The only reason is that I'm lazy about trying out the plan B, or plan C, D, or E, etc., such as replacing the whole python environment of Colab, applying TPU Research Cloud (TRC) program, or convincing my boss to use paid resources.
I understand perfectly, and I absolutely agree: the simpler the better! 

Concerning the error, I had to install an even newer version and had to correct this warning. I'm making a PR just after this one is merged to share this (very minor) change. :blush:

pyproject.toml
```diff
@@ -1,3 +1,28 @@
...
+
+[tool.poetry.dependencies]
+python = ">=3.7.11, <3.10"
+torch = "1.8.1"  # 1.9 also works for Colab TPU, but not for GPU/CPU; the bugfix of 1.8.2 seems irrelevant for now.
+hydra-core = ">=1.1,<1.2"  # Poetry will likely update it to 1.1.1, but it is probably ok and good for bugfixes.
+wandb = ">=0.10.32,<1"  # Poetry will likely update it to 0.12.1, but it is probably ok and good for bugfixes.
+transformers = "^4.6.0"  # Poetry will likely update it to 4.10.0, but it is probably ok and good for bugfixes.
+accelerate = ">=0.4.0,<1"  # We may want to use 0.5.0 in the near future
+datasets = {extras = ["streaming"], version = "^1.11.0"}
```
tianjianjiang, line 18: The semi-final changes for review.

requirements.txt
```diff
@@ -1,6 +1,6 @@
...
-wandb==0.10.30
-transformers==4.6.0
-accelerate==0.3.0
-datasets==1.11.0
+torch==1.8.1  # 1.9 also works for Colab TPU, but not for GPU/CPU; the bugfix of 1.8.2 seems irrelevant for now.
+hydra_core>=1.1,<1.2  # pip will likely update it to 1.1.1, but it is probably ok and good for bugfixes.
+wandb>=0.10.32,<1  # pip will likely update it to 0.12.1, but it is probably ok and good for bugfixes.
+transformers>=4.6.0,<5  # pip will likely update it to 4.10.0, but it is probably ok and good for bugfixes.
+accelerate>=0.4.0,<1  # We may want to use 0.5.0 in the near future
+datasets[streaming]>=1.11.0,<2
```
tianjianjiang, line 6: Same as https://github.com/bigscience-workshop/metadata/pull/19/files#r702606374
See https://python-poetry.org/docs/dependency-specification/ for translating Poetry's `^` to pip requirements.txt.

timoschick (approved):

SaulLu (approved):
Thanks @tianjianjiang !

PR #21, opened 2021-08-22
add OSCAR Entity_Tags 10_000 dataset

manandey:


timoschick (approved):
Looks good to me 👍 
For moving to bigger datasets, we'll have to think of other ways of sharing datasets (e.g., using gcp / huggingface datasets / ...), but I think it's very useful to have some medium size datasets directly in this repo.

manandey:
> Looks good to me 👍
> For moving to bigger datasets, we'll have to think of other ways of sharing datasets (e.g., using gcp / huggingface datasets / ...), but I think it's very useful to have some medium size datasets directly in this repo.

Sure, for the next set of datasets, I will try to add them in Huggingface Datasets.

SaulLu:
Thanks a lot for all your work @manandey 

I think that it is a good idea to store the dataset on the hub especially as there will soon (early this week) be a handy new feature for downloading datasets from the hub. :stars: 

I'll also just take the opportunity to share that I'm personally not a big fan of putting datasets in the code repo for 2 reasons: 1) the first is that a dataset is something that takes up space (and for example, we can't preview the file you just add on github) and 2) the second is that with JZ we have limits on the number of files for each of our partitions and we tend to try to store the code and datasets in a different partition.

manandey:
> Thanks a lot for all your work @manandey
> 
> I think that it is a good idea to store the dataset on the hub especially as there will soon (early this week) be a handy new feature for downloading datasets from the hub. 🌠
> 
> I'll also just take the opportunity to share that I'm personally not a big fan of putting datasets that goes beyond the example in the code repo for 2 reasons: 1) the first is that a dataset is something that takes up space (and for example, we can't preview the file you just add on github because it's too big) and 2) the second is that with JZ we have limits on the number of files for each of our partitions and we tend to try to store the code and datasets in a different partition. 🙂

Sure, Thanks a lot @SaulLu for the inputs! I will close this PR and store the dataset on the hub itself then.

PR #23, opened 2021-08-26
Remove global metadata separator when there is only local metadata

### LINKED ISSUE #22: Add metadata feature - Behavioral question
I was wondering if the output of the `add_metadata_and_chunk_examples` function was in the desired format. 

When we want to add only local metadata, the text example always begin as follows: ` |||>Releas`. Is it right to start with the separator even if you don't add global metadata?

SaulLu:
Fix #22 

In this PR, I propose to remove the `global_metadata_sep` when not adding global metadata but only local ones.

I changed the test accordingly :slightly_smiling_face:


timoschick (approved):
I had somehow completely overlooked this... thank you very much for taking care of it 👍

PR #26, opened 2021-09-06
[WIP] Changing the order of local tags

SaulLu:


tests/test_metadata_utils.py
```diff
@@ -133,6 +204,16 @@ def test_add_no_metadata_and_chunk_examples(self):
         for example in mapped_ds:
             self.assertTrue(all(not x for x in example["metadata_mask"]))
 
+    def test_add_html_tags(self):
+        cfg = DataConfig()
+        cfg.metadata_list = ["html"]
+        PROCESSORS["html"] = HtmlProcessor
+
+        text1, mask1 = add_local_metadata_to_text(self.examples[3], cfg)
+        target_text = '<a>useless text </a><div><a><div><div></div></div></a></div><h1><i>The Walking Dead</i> (season 8)</h1>\n'
```
SaulLu, line 213: Before the change I propose in this PR the result was here: 
```
<a>useless text </div></a></div></a></div><h1><i><div><a><div><div>The Walking Dead</i> (season 8)</h1>
```

tianjianjiang:
LGTM!

tianjianjiang (approved):
LGTM!

tianjianjiang:
> As a side note, with the current metadata format, one level of information is lost for HTM tags. Indeed, when two tags have the same index for their opening and closing tag, we don't know which one is the parent.

Just curious, is this (same index) a limitation of the source dataset?

SaulLu:
> Just curious, is this (same index) a limitation of the source dataset?

No no, I compile the dataset myself from [Natural Questions](Natural Questions) and put it in the format proposed [here](https://github.com/bigscience-workshop/metadata#example).

tianjianjiang:
> > Just curious, is this (same index) a limitation of the source dataset?
> 
> No no, I compile the dataset myself from [Natural Questions](https://ai.google.com/research/NaturalQuestions) and put it in the format proposed [here](https://github.com/bigscience-workshop/metadata#example).

Ah, I see. I misunderstood. Thank you for the clarification!
I suppose if we truly want to resolve the nested-tag-situation then there may be a bit too much work for your HTML parser?

SaulLu:
> I suppose if we truly want to resolve the nested-tag-situation then there may be a bit too much work for your HTML parser?
:slightly_smiling_face: 

> I suppose if we truly want to resolve the nested-tag-situation then there may be a bit too much work for your HTML parser?
On the parser side I think it's really feasible, what's more blocking in my opinion is where to store the information in the `jsonl` format we have defined and how to use it in the `add_local_metadata_to_text` method without it being something specific to HTML.

tianjianjiang:
> what's more blocking in my opinion is where to store the information in the `jsonl` format we have defined and how to use it in the `add_local_metadata_to_text` method without it being something specific to HTML.

I see. Funny you should mention that, because my first comment was gonna be "having a new property or even structure of jsonl for it" but then I didn't send it for exactly the same reason. lol

That being said, I wonder if entities will also need that kind of nested construct.

timoschick:
Hi @SaulLu, as discussed yesterday, I fully agree that this is an important issue that needs to be fixed. However, I'm not sure I fully understand how your proposed solution works - would you perhaps have 15 minutes or so sometime tomorrow (ideally, between 15:00 and 17:00 CEST) to go through this pull request with me? :)

SaulLu:
Summary of an offline discussion: unfortunately this PR does not solve all the concerns about the order of the local HTML metadata to be added. 

For example, @timoschick has rightly shown that `<tr><th>elephant</th><th></th></tr>` would be transformed into `<tr><th>elephant</th></tr><th></th>` with this PR (which is still not correct).

Another pathological case is:
```
<table>
  <tr>
    <th>event_id</th>
    <th>year</th>
    <th>month</th>
  </tr>
  <tr>
    <td>idghtu</td>
    <td>1998</td>
    <th>may</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>
```

I'm converting this PR into a draft, for now, the time to think about another workaround.

PR #27, opened 2021-09-06
Improve `datasets` capabilities

SaulLu:
In this PR I propose several changes that affect the dataset part of our script.

First, datasets uses caches. Caches are notably generated by `map` functions and their name depends on the hash of the function's parameters. `args` unfortunately had a non-deterministic hash, I propose to solve that.

Next, I propose two new arguments: `extension` and `map_batch_size`. The first is to be able to bring more generality than the extension guessed automatically by the code. The other argument is to adjust the batch size used by the map function of `datasets`. With a small batch size it is possible to reduce the risk of OOM (even if I still have some on big datasets) but for some operations (the `without_metatdata` script) the batch size still matters if the dataset examples are small (in this case we concatenate the batch examples to make a big example).

Finally I added logs to get more feedback when a problem occurs (like an OOM error).

### DIFF
--- a/bsmetadata/experiments/with_metadata.py
+++ b/bsmetadata/experiments/with_metadata.py
@@ -122,6 +122,7 @@ def get_dataloaders(tokenizer, args):
     tmp_data_args.overwrite_cache = False
     tmp_data_args.per_device_eval_batch_size = 2
     tmp_data_args.per_device_train_batch_size = 2
+    tmp_data_args.map_batch_size = 1
 
     # First we pre-process our text and metadata
     datasets = datasets.map(
@@ -131,7 +132,7 @@ def get_dataloaders(tokenizer, args):
         load_from_cache_file=not args.overwrite_cache,
         desc="Pre-process the text and metadata to create new samples",
         remove_columns=column_names,
-        batch_size=1, # To avoid out of memory issues
+        batch_size=args.map_batch_size,
     )
     logger.info("Add metadata and chunk examples finished")
 
@@ -147,7 +148,7 @@ def get_dataloaders(tokenizer, args):
         num_proc=args.preprocessing_num_workers,
         load_from_cache_file=not args.overwrite_cache,
         desc="Create labels column",
-        batch_size=1, # To avoid out of memory issues
+        batch_size=args.map_batch_size,
     )
     logger.info("Creating labels column finished")
 
--- a/bsmetadata/experiments/with_metadata.py
+++ b/bsmetadata/experiments/with_metadata.py
@@ -53,7 +53,7 @@ def get_dataloaders(tokenizer, args):
     logger.info(f"Start to load dataset, the result will be cached at {config.HF_DATASETS_CACHE}")
     if args.dataset_name is not None:
         logger.info(
-            "Downloading and loading with arguments: "
+            "Downloading with arguments: "
             f"dataset_name={args.dataset_name}, "
             f"dataset_config_name={args.dataset_config_name}, "
             f"data_files={data_files}, "
@@ -131,6 +131,7 @@ def get_dataloaders(tokenizer, args):
         load_from_cache_file=not args.overwrite_cache,
         desc="Pre-process the text and metadata to create new samples",
         remove_columns=column_names,
+        batch_size=1, # To avoid out of memory issues
     )
     logger.info("Add metadata and chunk examples finished")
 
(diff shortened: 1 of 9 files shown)


bsmetadata/experiments/with_metadata.py
```diff
@@ -40,88 +42,119 @@ def get_dataloaders(tokenizer, args):
     #
     # In distributed training, the load_dataset function guarantees that only one local process can concurrently
     # download the dataset.
+    data_files = {}
+    if args.train_file is not None:
+        data_files["train"] = args.train_file
+    if args.validation_file is not None:
+        data_files["validation"] = args.validation_file
```
SaulLu, line 49: I put this block up because you can also specify data_files to be considered in a hub repo using the new `datasets` library feature (currently on master).

bsmetadata/input_pipeline.py
```diff
@@ -50,6 +50,7 @@ class DataConfig:
     cache_dir: Optional[str] = field(
         default=None, metadata={"help": "Where do you want to store the pretrained models downloaded from s3?"}
     )
+    extension: Optional[str] = field(default=None, metadata={"help": "the file extension of the dataset"})
```
SaulLu, line 53: I added this argument as the `json` script also work for `json.gz` files

cccntu (approved):
LGTM! It's great that you found the hash issue and fixed it. 

Side note: `omegaconf.to_object` is a newly added feature that I am not aware of. Looks very useful.

bsmetadata/experiments/with_metadata.py
```diff
@@ -40,88 +41,122 @@ def get_dataloaders(tokenizer, args):
...
+    tmp_data_args.overwrite_cache = False
+    tmp_data_args.per_device_eval_batch_size = 2
+    tmp_data_args.per_device_train_batch_size = 2
+    tmp_data_args.map_batch_size = 1
 
     # First we pre-process our text and metadata
-    lm_datasets = raw_datasets.map(
-        functools.partial(add_metadata_and_chunk_examples, tokenizer=tokenizer, cfg=args),
+    datasets = datasets.map(
+        functools.partial(add_metadata_and_chunk_examples, tokenizer=tokenizer, cfg=tmp_data_args),
```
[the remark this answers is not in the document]
cccntu: Optional, could be done in another patch.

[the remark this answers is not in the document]
SaulLu: I really like this suggestion!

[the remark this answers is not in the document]
SaulLu: I included it in the commit I just pushed. Could you tell me what you think? :slightly_smiling_face:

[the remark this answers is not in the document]
cccntu: Thanks fo taking care of it. It looks great to me. 😃

bsmetadata/experiments/with_metadata.py
```diff
@@ -40,88 +40,113 @@ def get_dataloaders(tokenizer, args):
...
     )
+    logger.info("Creating labels column finished")
+
+    train_dataset = datasets["train"]
+    val_dataset = datasets["validation"]
 
-    train_dataset = lm_datasets["train"]
-    val_dataset = lm_datasets["validation"]
+    logger.info(f"  Num train examples = {len(train_dataset)}")
+    logger.info(f"  Num validation examples = {len(val_dataset)}")
```
tianjianjiang, line 149: I love this change and just wonder if any of you also like to see the numbers before tokenization and block-grouping/chunking.

bsmetadata/input_pipeline.py
```diff
@@ -1,38 +1,21 @@
...
+from bsmetadata.metadata_utils import MetadataConfig
 
 
 @dataclass
 class DataConfig:
+    metadata_config: MetadataConfig = field(
+        default_factory=lambda: MetadataConfig(), metadata={"help": "The configuration related to the metadata."}
+    )
     experiment: str = field(default="sample", metadata={"help": "The name of the experiment."})
     per_device_eval_batch_size: int = field(
```
tianjianjiang, line 13: Just a side note. On Colab I noticed that I can use `8` for eval.

SaulLu: That a good thing to know!

bsmetadata/experiments/with_metadata.py
SaulLu: That is a good idea! :slightly_smiling_face: I'd like this addition too

PR #28, opened 2021-09-06
refactor: eval tqdm without newline

tianjianjiang:
Otherwise eval newlines will be flooding the output of (Colab) notebook.

https://tqdm.github.io/docs/tqdm/#__init__
> **position**: _int_, _optional_
> Specify the line offset to print this bar (starting from 0) Automatic if unspecified. Useful to manage multiple bars at once (eg, from threads).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Code quality
+name: Test
 
 on:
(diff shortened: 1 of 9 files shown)


SaulLu (approved):
Looks great! Thanks

PR #31, opened 2021-09-15
Add eval and save options

SaulLu:
## What this PR do?
I propose to add some options to adjust the number of evaluations and the saving of checkpoints. Let me know what you think :blush:


tianjianjiang (approved):
LGTM and just curious why there are many `+1`

bsmetadata/train.py
```diff
@@ -156,10 +195,26 @@ def main(args: CFG) -> None:
     else:
         args.num_train_epochs = math.ceil(args.max_train_steps / num_update_steps_per_epoch)
 
-    if args.num_eval < 1:
+    if args.evaluation_strategy == IntervalStrategy.EPOCH:
+        if args.eval_num_per_epoch < 1:
+            eval_per_n_step = args.max_train_steps + 1
```
SaulLu, line 200: @tianjianjiang, are you referring to this kind of `+1` ? :slightly_smiling_face:

tianjianjiang: @SaulLu Yes~

SaulLu: I just kept what was done before, and it seems to me that this has the effect of not doing an evaluation. :slightly_smiling_face:

timoschick (approved):
LGTM 😊

bsmetadata/train.py
```diff
@@ -156,10 +195,26 @@ def main(args: CFG) -> None:
...
+    else:  # IntervalStrategy.NO
         eval_per_n_step = args.max_train_steps + 1
-    else:
-        eval_per_n_step = args.max_train_steps // args.num_eval
+
+    if args.save_strategy == IntervalStrategy.EPOCH:
+        if args.save_num_per_epoch < 1:
+            save_per_n_step = args.max_train_steps + 1
+        else:
+            save_per_n_step = args.max_train_steps // args.save_num_per_epoch
```
cccntu, line 212: (nit)
This can be re-written to:
```
if (args.save_strategy == IntervalStrategy.EPOCH 
   and args.save_num_per_epoch >= 1):
    save_per_n_step = args.max_train_steps // args.save_num_per_epoch
```
so there is fewer lines and fewer `+1`

bsmetadata/train.py
cccntu: Can confirm, I did it in the first place, I should've added a comment `# will never eval`. 🙃

bsmetadata/train.py
```diff
@@ -156,10 +195,26 @@ def main(args: CFG) -> None:
     else:
         args.num_train_epochs = math.ceil(args.max_train_steps / num_update_steps_per_epoch)
 
-    if args.num_eval < 1:
+    if args.evaluation_strategy == IntervalStrategy.EPOCH:
+        if args.eval_num_per_epoch < 1:
+            eval_per_n_step = args.max_train_steps + 1
+        else:
+            eval_per_n_step = args.max_train_steps // args.eval_num_per_epoch
```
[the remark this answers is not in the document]
SaulLu: You are absolutely right! I pushed the change! Thank you for your careful proofreading

bsmetadata/train.py
SaulLu: It's a good idea, I pushed the change

bsmetadata/train.py
```diff
@@ -212,29 +285,35 @@ def evaluate(eval_dataloader):
...
+                save_dict = {
+                    "epoch": epoch + 1,
+                    "state_dict": accelerator.unwrap_model(model).state_dict(),
+                    "optimizer": optimizer.state_dict(),
+                    "scheduler": scheduler.state_dict(),
+                }
+                torch.save(
+                    save_dict,
+                    save_path,
+                )
```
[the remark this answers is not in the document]
SaulLu: Done

SaulLu:
Thank you very much for your reviews! :hugs: 

@cccntu,
> What's reason you use IntervalStrategy instead of something like this:
[...]
I think IntervalStrategy make the code more explicit. thinking

I think you've spotted it, the main reason is I think to simplify the reading of the code and also, to a second extent, the choice of the training parameters (no need to make sure that the `X` argument overrides the `Y` argument) :slightly_smiling_face: 
(and the third reason is that `transformers`' training arguments work like this and I find it convenient)

> Do you think the code above should use IntervalStrategy, too?
I don't really have an opinion! It's true that it could also bring the same "ease" of reading ifs conditions but on the other hand we only have two possible states for the `update_strategy` (`steps` and `epochs`). Would you have a preference?

cccntu (changes requested):
@SaulLu 
Thank you for explaining it to me. I don't have a preference, so I think we can leave it as is.

One last thing and it's good to go.

bsmetadata/train.py
```diff
@@ -42,9 +47,43 @@ class CFG:
...
+    save_steps: int = field(default=500, metadata={"help": "Save checkpoint every X update steps."})
+    # save_total_limit: Optional[int] = field(
+    #     default=None,
+    #     metadata={
+    #         "help": (
+    #             "Limit the total amount of checkpoints."
+    #             "Deletes the older checkpoints in the output_dir. Default is unlimited checkpoints"
+    #         )
+    #     },
+    # )  # might be usefull to had, especially for google colab experiment
```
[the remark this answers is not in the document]
SaulLu: I have removed this comment in the new commit :slightly_smiling_face:

bsmetadata/train.py
```diff
@@ -185,9 +234,27 @@ def evaluate(eval_dataloader):
...
+            metrics = evaluate(eval_dataloader)
+            metrics_logger.log({key: metrics})
+        logger.info("Evaluation finished")
+
+    if not args.do_train:
+        return
+
+    logger.info("Start training")
+    logger.info(f"  Evaluation will be done every {eval_per_n_step} steps")
+    logger.info(f"  Saving will be done every {save_per_n_step} steps")
```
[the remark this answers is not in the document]
SaulLu: Good idea! I've changed the logging message in the last commit

Issue #36, opened 2021-09-21
wip: feat: Add special tokens to request (or not) metadata generation


SaulLu:
# Add special tokens to request (or not) metadata generation

## Feature details

To be completed

SaulLu:
I really like your proposal too! I just have a question with your proposal, what rule should we follow for the order of the metadata in the prefix (also the order in `metadata_list`)?

Going back to my proposal, I had actually thought about adding special tokens that seem easy to interpret for the model. I have a feeling that the second proposal might be slightly more complicated to interpret (as the global metadata descriptions have a very variable length). 

In any case, I don't have a strong opinion! Another possibility could be to isolate at the beginning only special tokens for local metadata (and also to implement and test both options)

tianjianjiang:
My two cents: I usually do this like what fast.ai does
```python
tokenizer.add_special_tokens(["xxxglobal",  "xxxlocal"])
```
`xxxglobal html xxxglobal url ...://... xxxglobal timestamp ... ||| this is a main text that also says html, url, timestamp as non-special tokens`

manandey:
Hi @SaulLu, I had some difficulty in understanding the generated sample in this example:

> ### Generated example 2
> With the arguments:
> 
> ```
> metadata_list = ["url", "timestamp", "entity", "html"] <- change here
> special_token_for_metadata_generation = True
> special_token_for_metadata_generation_sep = " ||| "
> metadata_probability = 0
> ```
> 
> Generated sample:
> 
> ```
> url timestamp html ||| url: https://www.bbc.com/sport/live/olympics/50974152 | timestamp: 2018-12-10T13:45:00.000Z ||| <div class:summary> the  2018 winter Olympic Games was held between 9 and 25 February 2018 in South Korea</div>
> ```

SaulLu:
> My two cents: I usually do this like what fast.ai does 
```tokenizer.add_special_tokens(["xxxglobal",  "xxxlocal", "|||"])```
`xxxglobal html xxxglobal url ...://... xxxglobal timestamp ... ||| this is a main text that also says html, url, and timestamp as non-special tokens`

[...]

PR #38, opened 2021-09-23
wip: feat: add special tokens to request (or not) metadata generation

SaulLu:
Solve issue #36

### DIFF
--- a/bsmetadata/metadata_processors.py
+++ b/bsmetadata/metadata_processors.py
@@ -36,9 +36,23 @@ class MetadataConfig:
     metadata_probability: float = field(
         default=1, metadata={"help": "The probability of adding metadata to an input example."}
     )
+    add_special_token_for_metadata_generation: bool = field(
+        default=True,
+        metadata={
+            "help": "If True, some special tokens are added at the begining of the sample to indicate the type of "
+            "metadata added in the sample. The special tokens used are equal to the string used in `metadata_list`"
+        },
+    )
     global_metadata_sep: str = field(
         default=" |||",
         metadata={"help": "The character sequence that is used to separate all global metadata from the actual text."},
+    )
+    special_token_for_metadata_generation_sep: str = field(
+        default=" |||",
+        metadata={
+            "help": "The character sequence that is used to separate the special tokens controlling the generation of "
+            "metadata from (eventually) the global metadata and the actual text."
+        },
     )
     max_seq_len: int = field(
         default=512, metadata={"help": "The maximum number of tokens to use for each training chunk."}


tests/test_metadata_utils.py
```diff
@@ -1641,6 +1679,102 @@ def test_add_metadata_and_chunk_examples_with_true_processor(self):
...
+        self.assertEqual(
+            self.tokenizer.decode(mapped_ds[1]["input_ids"]),
+            " url html entity ||| url: https://en.wikipedia.org/wiki/Apple ||| An <b>apple [[Malus domestica]]</b> is an edible fruit produced by an <a></a><b class:level1><b class:level2><i class:level3><a></a>apple</i> tree</b></b> (Malus domestica).<|endoftext|>",
+        )
+        self.assertEqual(
+            self.tokenizer.decode(mapped_ds[2]["input_ids"]),
+            " url html entity ||| url: https://en.wikipedia.org/wiki/Apple ||| An <b>apple [[Malus domestica]]</b> is an edible fruit produced by an <a></a><b class:level1><b class:level2><i class:level3><a></a>apple</i> tree</b></b> (Malus domestica).<|endoftext|>",
+        )
+
+        # fmt: off
```
SaulLu, line 1724: Here I added a feature to ignore the formatting between `# fmt: off` and `# fmt: on`. I found it easier to visualize the target examples. What do you think about it?

bsmetadata/metadata_utils.py
```diff
@@ -137,9 +175,11 @@ def create_global_metadata_prefix(example: Dict[str, Any], cfg: MetadataConfig)
         processor = PROCESSORS.get(key, MetadataProcessor)(cfg)
         processed_metadata[key] = processor.process_global(metadata)
 
-    sorted_metadata = [processed_metadata.get(md, None) for md in cfg.metadata_list]
-    sorted_metadata = [md for md in sorted_metadata if md is not None]
-    return cfg.metadata_sep.join(sorted_metadata) + cfg.global_metadata_sep if sorted_metadata else ""
```
SaulLu, line 142: I changed the behavior of this function by not adding the separator between the global metadata string and the text string (`cfg.metadata_global_sep`) in the `create_global_metadata_prefix` method. This way, both types of separators (`cfg.metadata_global_sep` and `metadata_special_token_for_generation_sep`) are added within the `add_metadata_and_chunk_examples` method.

tests/test_metadata_utils.py
```diff
@@ -1641,6 +1679,102 @@ def test_add_metadata_and_chunk_examples_with_true_processor(self):
             "url: https://en.wikipedia.org/wiki/Apple ||| An <b>apple [[Malus domestica]]</b> is an edible fruit produced by an <a></a><b class:level1><b class:level2><i class:level3><a></a>apple</i> tree</b></b> (Malus domestica).<|endoftext|>",
         )
 
+    def test_add_metadata_and_chunk_examples_with_true_processor_and_metadata_special_tokens(self):
+        cfg = MetadataConfig()
+        cfg.metadata_list = ["url", "timestamp", "html", "entity"]
+        cfg.max_seq_len = 85
+        cfg.metadata_probability = 1
+        cfg.metadata_add_special_token_for_generation = True
+        cfg.metadata_special_token_for_generation_start_seq = " "
```
SaulLu, line 1688: adding this space allows metadata types to be always tokenised in the same way:
`" html entity"` -> `['Ġhtml', 'Ġentity']`
`" entity html"` -> `['Ġentity', 'Ġhtml']`

tianjianjiang (commented):
I'm not 100% sure I got the idea, and yet I imagine that there are two kinds of special token in our use case:

1. separator
2. metadata type indicator

IMO, the PR is totally fine with (2), but for (1) I guess we can use `tokenizer.add_special_tokens()` such that we don't have to have tricky whitespace as prefix/suffix.

bsmetadata/metadata_utils.py
```diff
@@ -126,20 +164,28 @@ def create_global_metadata_prefix(example: Dict[str, Any], cfg: MetadataConfig)
...
     processed_metadata = {}
+    global_metadata_special_tokens = []
     for metadata in example["metadata"]:
         key, type_ = metadata["key"], metadata["type"]
         if type_ != "global" or key not in cfg.metadata_list:
             continue
 
+        if key not in global_metadata_special_tokens:
+            global_metadata_special_tokens.append(key)
+
```
cccntu, line 180: IIUC, this `if` will make the new prefix only contain local metadata types, right?

cccntu (approved):
LGTM. Thank you for adding this!

SaulLu:
@tianjianjiang 
> IMO, the PR is totally fine with (2), but for (1) I guess we can use tokenizer.add_special_tokens() such that we don't have to have tricky whitespace as prefix/suffix.

You make a very good point! Here I wanted to leave the possibility to work with the gpt2 tokenizer as is. Indeed the addition of new tokens will require a small modification in the embedding layer of the model and thus introduce some untrained parameters if we continue to train the `gpt2` checkpoint.

But in the long run I agree that adding special tokens is the best solution for managing these spaces.

bsmetadata/metadata_utils.py
SaulLu: outdated

SaulLu:
@timoschick, @cccntu and @tianjianjiang , for your information I have updated this PR to implement the other format we agreed on :blush:

cccntu (approved):
LGTM. The logic is short now! 🥳

tests/test_metadata_utils.py
```diff
@@ -1641,6 +1647,166 @@ def test_add_metadata_and_chunk_examples_with_true_processor(self):
...
+            batched=True,
+            remove_columns=ds.column_names,
+            load_from_cache_file=False,
+        )
+
+        self.maxDiff = None
+
+        self.assertEqual(
+            self.tokenizer.decode(mapped_ds[0]["input_ids"]),
+            " url: https://en.wikipedia.org/wiki/Apple | html | entity ||| An <b>apple [[Malus domestica]]</b> is an edible fruit produced by an <b class:level1><i class:level2>apple</i> tree</b> (Malus domestica).<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>",
```
tianjianjiang, line 1680:
```suggestion
            f" url: https://en.wikipedia.org/wiki/Apple | html | entity ||| An <b>apple [[Malus domestica]]</b> is an edible fruit produced by an <b class:level1><i class:level2>apple</i> tree</b> (Malus domestica).{'<|endoftext|>'*17}",
```

I guess that the number of `<|endoftext|>` is also an implicit part of the assertion.

---
> ℹ️ Mike's reviewing oath 🖖

tests/test_metadata_utils.py
```diff
@@ -1641,6 +1647,166 @@ def test_add_metadata_and_chunk_examples_with_true_processor(self):
...
+        ds = Dataset.from_dict(ds_dict)
+
+        mapped_ds = ds.map(
+            functools.partial(add_metadata_and_chunk_examples, tokenizer=self.tokenizer, cfg=cfg),
+            batched=True,
+            remove_columns=ds.column_names,
+            load_from_cache_file=False,
+        )
+
+        self.maxDiff = None
```
tianjianjiang, line 1676: I'm sorry that I failed to notice this sooner when it was introduced previously. Since it was also [set to None during the `setUp()`](https://github.com/bigscience-workshop/metadata/blob/master/tests/test_metadata_utils.py#L26), would this be a special treatment for certain test functions?

tianjianjiang (approved):
LGTM with two comments. Thank you for the nicely done PR as usual!

(Pardon me for the intrusion. I took the liberty of changing the PR description by using GitHub syntax of issue association.)

---
> ℹ️ Mike's reviewing oath 🖖

tests/test_metadata_utils.py
SaulLu: Suggestion commited! Thank you :)

tests/test_metadata_utils.py
SaulLu: Suggestion commited! Thank you :)

tests/test_metadata_utils.py
SaulLu: Honestly, I don't know why I needed this argument is here. If it's there it's probably because I needed it to see all the output when I had an error in the tests, then it's possible that when I added it there was not yet the `self.maxDiff` set to None in the `setUp()` (and that it arrived later in a merge with the `master` branch).

PR #41, opened 2021-09-28
Add different settings in which entities can be used

manandey:
Hi @timoschick ,

As discussed earlier, this PR allows us to use entities at the beginning and end of each paragraph. eg.

**END:**
`"It was a brilliant first round. You have to break down the Cuban's rhythm you can't let them get into rhythm. <ENTITY_CHAIN> [[Cuban]] </ENTITY_CHAIN> \nThe risk with that is Yafai has got to go him <ENTITY_CHAIN> [[Galal Yafai]] </ENTITY_CHAIN>."
`
**BEG:**
` "<ENTITY_CHAIN> [[Cuban]] <SUMMARY> It was a brilliant first round. You have to break down the Cuban's rhythm you can't let them get into rhythm. \n<ENTITY_CHAIN> [[Galal Yafai]] <SUMMARY> The risk with that is Yafai has got to go him."`

Can you kindly have a look into it and let me know if any changes need to be done.


bsmetadata/metadata_utils.py
```diff
@@ -207,6 +207,25 @@ def _collate_metadata(metadata_list: List[dict], cfg: MetadataConfig):
...
+
+    Returns:
+        The start_text merged with the semi_local_metadata_sep in required order
+    """
+    if (("[[" or "]]") in start_text) and cfg.entity_setting == "beg":
+        return f"<ENTITY_CHAIN>{start_text} </ENTITY_CHAIN> "
+    elif (("[[" or "]]") in start_text) and cfg.entity_setting == "end":
+        return f" <ENTITY_CHAIN>{start_text} </ENTITY_CHAIN>"
+    else:
+        return start_text
```
shanyas10, line 226:
```suggestion
    beg_space = "" if cfg.entity_setting == "beg" else " "
    end_space = "" if cfg.entity_setting == "end" else " "
    if ("[[" or "]]") in start_text:
        return f"{beg_space}<ENTITY_CHAIN>{start_text} </ENTITY_CHAIN>{end_space}"
    return start_text
```

manandey: Thanks! @shanyas10

bsmetadata/metadata_utils.py
```diff
@@ -182,15 +182,26 @@ def _collate_metadata(metadata_list: List[dict], cfg: MetadataConfig):
...
+            metadata_dict_idx[metadata_node.char_end_idx][metadata_node.relative_end_pos] = end_text
 
     for absolute_idx, value in metadata_dict_idx.items():
         pos_sorted = sorted(list(value.keys()))
         local_metadata = ""
         for pos in pos_sorted:
             local_metadata += metadata_dict_idx[absolute_idx][pos]
+
+        # We add here a local special token if needed around the metadata list of a type if needed
+        if local_metadata and metadata_list[0]["key"] in cfg.local_metadata_special_token_start:
```
shanyas10, line 198: @manandey I think we should also check if `cfg.local_metadata_special_token_start` is non null before iterating through it. Same for `special_token_end`

bsmetadata/metadata_utils.py
```diff
@@ -183,14 +183,33 @@ def _collate_metadata(metadata_list: List[dict], cfg: MetadataConfig):
...
+            local_special_token_start = cfg.local_metadata_special_token_start[metadata_list[0]["key"]]
+            local_metadata = f"{local_special_token_start}{local_metadata}"
+        if (
+            cfg.local_metadata_special_token_end
+            and local_metadata
+            and metadata_list[0]["key"] in cfg.local_metadata_special_token_end
+        ):
+            local_special_token_end = cfg.local_metadata_special_token_end[metadata_list[0]["key"]]
+            local_metadata = f"{local_metadata}{local_special_token_end}"
+
```
tianjianjiang, line 212: Two minor questions:

1. Just curious, is it possible that a `metadata_list[0]["key"]` can be in both `cfg.local_metadata_special_token_start` and  `cfg.local_metadata_special_token_end`? I understand that we usually have an opening-closing tag pair for those special tokens, so technically it should never happen. If that is always the case, then perhaps we can have one `cfg.local_metadata_special_token_something` and then control the opening/closing by `cfg,entity_setting`, if I read the code correctly.
2. I guess we don't have to deal with nested `ENTITY_CHAIN`s because impossible by design?

---
> ℹ️ Mike's reviewing oath 🖖

tianjianjiang (approved):
@manandey Thank you for the PR! The feature looks intriguing. I don't have any critical opinion about it, so just virtually approving this with some comments. Please still ask others for a real review. 🙏

bsmetadata/metadata_utils.py
```diff
@@ -183,14 +183,33 @@ def _collate_metadata(metadata_list: List[dict], cfg: MetadataConfig):
         assert metadata_node.relative_start_pos not in metadata_dict_idx[metadata_node.char_start_idx]
         assert metadata_node.relative_end_pos not in metadata_dict_idx[metadata_node.char_end_idx]
 
-        metadata_dict_idx[metadata_node.char_start_idx][metadata_node.relative_start_pos] = start_text
-        metadata_dict_idx[metadata_node.char_end_idx][metadata_node.relative_end_pos] = end_text
+        if start_text:
+            metadata_dict_idx[metadata_node.char_start_idx][metadata_node.relative_start_pos] = start_text
+        if end_text:
+            metadata_dict_idx[metadata_node.char_end_idx][metadata_node.relative_end_pos] = end_text
```
timoschick, line 189: Why do we need these `if start_text:` and `if end_text:` checks here?

bsmetadata/metadata_utils.py
timoschick, line 212: I'm not sure if we've discussed this earlier and I just forgot about it, but this seems to be very specific to entity metadata. Wouldn't it be better then to remove this logic from the main processing and instead add it to the custom metadata processor? Or is there some other kind of metadata that would also require this?

bsmetadata/metadata_utils.py
SaulLu: Let me answer this question as I was the one who advised @manandey  to make this change (and we discussed it outside github - not a good practice indeed :disappointed: ).

These checks allow to use the same dataset to do experiments in `beg` mode and in `end` mode. So we don't have to rework the dataset upstream to do one or the other experiment

bsmetadata/metadata_utils.py
SaulLu: I also take the opportunity to share my thoughts on this point when we talked about it with Manan (I remember asking the exact same questions initially, it's funny!). By the way, the only trace on github of our exchange with Manan at the time is [this PR](https://github.com/manandey/metadata/pull/1) . :slightly_smiling_face: 

In the end, I was thinking that this framework could allow to do what you had proposed to do a while ago : frame local tags with tag detectors to be able to remove them. 

Cf, the notes of the 2021 Sep 17 meeting with the following example: 
```
use special token: This is a [BEGIN_MD type=html] <b> [END_MD] great [BEGIN_MD type=html] </b> [END_MD] apple [BEGIN_MD type=entity] {{Apple (fruit)}} [END_MD]
```

bsmetadata/metadata_utils.py
```diff
@@ -203,7 +203,7 @@ def _collate_metadata(metadata_list: List[dict], cfg: MetadataConfig):
             local_special_token_start = cfg.local_metadata_special_token_start[metadata_list[0]["key"]]
             local_metadata = f"{local_special_token_start}{local_metadata}"
         if (
-            cfg.local_metadata_special_token_start
+            cfg.local_metadata_special_token_end
```
SaulLu, line 206: Good catch :wink:

bsmetadata/metadata_processors.py
```diff
@@ -40,9 +40,21 @@ class MetadataConfig:
...
+    entity_setting: str = field(
+        default="normal",
+        metadata={"help": "The settings in which you want to use entites. Valid choices: (beg, end, normal)"},
+    )
     max_seq_len: int = field(
         default=512, metadata={"help": "The maximum number of tokens to use for each training chunk."}
     )
+    local_metadata_special_token_start: Optional[Dict[str, str]] = field(
+        default=None,
+        metadata={"help": "A dictionary whose keys correspond to a local metadata type and values ...."},
```
SaulLu, line 52: It seems to me that this description (that I had proposed to you, sorry :see_no_evil: ) is not finished. Would you mind finishing it?

bsmetadata/metadata_processors.py
```diff
@@ -40,9 +40,21 @@ class MetadataConfig:
...
+    )
+    local_metadata_special_token_end: Optional[Dict[str, str]] = field(
+        default=None,
+        metadata={"help": "A dictionary whose keys correspond to a local metadata type and values..."},
```
SaulLu, line 56: same here

SaulLu (commented):
Unless I'm mistaken, only a small bug has been fixed since the last time I reviewed your PR so everything looks good! Thank you @manandey ! 

One small point: I had opened for discussion a PR to show you the changes we could make in the code and had not finished writing the description of 2 arguments. I'm really sorry for that :pleading_face: . You'll find a comment below related to this point.

Issue #37, opened 2021-09-21
feat: resume training from a checkpoint


SaulLu:
# Feature request

It might be useful to be able to continue training left in a checkpoint, unless I am mistaken this is not a feature that is currently included

tianjianjiang:
Just a note: https://github.com/huggingface/accelerate/issues/95

PR #42, opened 2021-10-01
adding support to resume from a saved model (checkpoint)

shanyas10:
close: #37 (if this PR won't close the entire issue then "related: #37.")

In the current PR, the model along with tokenizer, configs and optimizer and scheduler states get saved at each checkpoint (based on the value of save_steps / save_num_per_epoch). 
 
 An extension to this could be to save only the best checkpoint till now (or have an option to support that on the user's wish). I can go ahead and implement that if that sounds good.


shanyas10:
#37

SaulLu:
Thank you so much for working on this feature @shanyas10 ! It really will 
be super useful.

The first question I have is 
I would have is: did you get to test this new setup? Especially in view of the comment shared by @tianjianjiang in the issue? From my point of view, it's really useful to check (even on a toy example) that everything goes as desired.

Then, I wonder if there is a particular reason why you save the model weights with `save_model_and_tokenizer` (L. 313) and the `state_dict` (L. 319). If I'm not mistaken, we'll have 2 saves (and the model weights are among the heaviest information to store). 

Finally, I think we can surely find a way to save the tokenizer only once (at the very beginning of the training for example) because the tokenizer will not change during the training and saving it each time will take a bit of computing time.

Idea: maybe we can use `save_model_and_tokenizer` juste one time at the first saving in order to save all the configs that will be usefull to reload the model ?

tianjianjiang:
Pardon me for the intrusion. I took the liberty of changing the PR description by using GitHub syntax of issue association.

---
> ℹ️ Mike's reviewing oath 🖖

bsmetadata/train.py
```diff
@@ -149,12 +173,16 @@ def main(args: CFG) -> None:
...
     tokenizer.pad_token = tokenizer.eos_token
     train_dataloader, eval_dataloaders = get_dataloaders(tokenizer, args.data_config)
 
     # get model
-    model = AutoModelForCausalLM.from_pretrained(args.model_name)
+    if not resumed_state:
+        model = AutoModelForCausalLM.from_pretrained(model_name)
+    else:
+        print("Loading model from checkpoint")
+        model = AutoModelForCausalLM.from_pretrained(resumed_state, from_pt=True)
```
cccntu, line 185: Does `from_pretrained` accept a state_dict (`resumed_state`)?
I don't think this is correct, but I did not try it.

I would `        model = AutoModelForCausalLM.from_pretrained(model_name)`
then `model.load_state_dict` if `resumed_state`

And put all of the if-then-load together in a single `if`

cccntu (commented):
Thank you for implementing this!
Some comments below.

bsmetadata/train.py
cccntu: Thank you. Yeah it does make more sense to put scheduler loading there, together with scheduler creation.

LGTM.

SaulLu (approved):
Thank you so much for making all these changes! Everything is good for me! :rocket:

timoschick (approved):
LGTM 👍

cccntu (approved):
LGTM.

PR #46, opened 2021-10-05
adding jz script for bigger model

shanyas10:
@SaulLu I've kept this the same as the one for toy example only made a change in the dataset. The dataset now consists of 15k training samples and 1500 dev


shanyas10:
Thanks @SaulLu ! Merging this

PR #49, opened 2021-10-15
Adding WebsiteMetadataProcessor to preprocessing_utils

shanyas10:
@timoschick raising this PR to understand if this is what is expected. Also, while annotating data myself, I had taken the top 100 keywords since they were most likely to have a wikipedia description. But since we aren't doing that here, chances are that some examples might not contain the description.   

Yet to write tests to the PR is not mergeable.


bsmetadata/preprocessing_utils.py
```diff
@@ -52,3 +55,63 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
     def _extract_timestamp_from_url(self, url: str) -> Optional:
         # This would have to be implemented.
         return None
+
+
+class WebsiteDescPreprocessor(MetadataPreprocessor):
+    """Metadata preprocessor for adding website description based on URLs."""
+
+    website_description_cache = {}
+    org_list = ["com", "co", "org", "go", "in"]
```
timoschick, line 64: What's the reason for using this specific set of top-level domains?

bsmetadata/preprocessing_utils.py
```diff
@@ -52,3 +55,63 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+            urls = [md["value"] for md in metadata if md["key"] == "url"]
+
+            if not urls:
+                continue
+
+            # Try to extract a website description from the given URL and add it to the metadata.
+            website_description = self._extract_website_desc_from_url(urls[0])
+
+            if website_description:
+                metadata.append({"key": "timestamp", "type": "global", "value": website_description})
```
timoschick, line 82: This should probably be something like `"key": "website_description"`

bsmetadata/preprocessing_utils.py
```diff
@@ -52,3 +55,63 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+
+        return examples
+
+    def _extract_website_desc_from_url(self, url: str) -> Optional:
+
+        domain = url.str.split("/")[2]  # e.g http://www.californialandcan.org/Plumas -> www.californialandcan.org
```
timoschick, line 88: This would fail for URLs that don't start with `http://`. I guess all URLs in C4 start with `http://`, but it would probably be good to be on the safe side here. Also, you may want to consider using some library like `urllib` (see https://docs.python.org/3/library/urllib.parse.html) for splitting URLs into components as this will take care of all unexpected edge cases for you.

bsmetadata/preprocessing_utils.py
```diff
@@ -52,3 +55,63 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+            if (keywords[1] not in self.org_list)
+            else keywords[0]
+        )  # extracting the keyword from domain e.g.  www.californialandcan.org -> californialandcan
+
+        if keyword not in self.website_description_cache:
+            self.website_description_cache[keyword] = self.extract_wiki_desc(keyword)
+
+        return self.website_description_cache[keyword]
+
+    def extract_wiki_desc(self, keyword: str) -> Optional:
```
timoschick, line 104: As JeanZ has no access to the internet, I think we need to first download all wiki descriptions as a preprocessing step.

shanyas10:
download_wiki_dump.sh -> Will clone the repo that contains the dump_db file for fetching wikipedia data
WebsiteMetadataProcessor -> needs to be initialized with the path to the above repo

Points for discussion:

* website_desc_utils L[19:27] is fetching the first two sentences of description for a given title. Splitting on ('. ') might not work for some cases e.g. "XYZ is a U.S. based company." would partition as "XYZ is a U.S" and "based company". While it worked for all the cases I tested for (around 10) but we can use a more robust method if any.
* Suggestions on how to test: Using the processor requires the download of the db file as a prerequisite. Any suggestions on how we can test the processor without the download? Or should we create a db file from a smaller wiki dump and use that as test data?

shanyas10:
@timoschick I see that the current commit contains some changes that are not a part of the processor. Please hold on the review till I revert them back.

bsmetadata/preprocessing_tools/website_desc_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    def fetch_wikipedia_title_from_keyword(self, keyword: str) -> str:
+        title = self.redirects_map.get(
+            keyword, keyword.split(".")[0].capitalize()
+        )  # fallback to default for cases where domain is not recognized. We'll try to hit the db with the exact keyword directly (e.g. rightmove.com -> Rightmove) Capitalizing since wikipedia titles are so
+        return title
+
+    def fetch_wikipedia_description_for_title(self, title: str) -> Optional:
+        try:
+            text = self.wiki_dump_db.get_paragraphs(title)[0].text
+            text = nltk.sent_tokenize(text)[0]  # Picking the first sentence
```
timoschick, line 25: Would it maybe make sense to remove information in brackets? For example, the first sentence in the Wikipedia article on Wikipedia itself is `Wikipedia (/ˌwɪkɪˈpiːdiə/ (listen) wik-ih-PEE-dee-ə or /ˌwɪki-/ (listen) wik-ee-) is a free content, multilingual online encyclopedia written and maintained by a community of volunteers through a model of open collaboration, using a wiki-based editing system.` At least in this case, I would think that the part in brackets (`/ˌwɪkɪˈpiːdiə/ (listen) wik-ih-PEE-dee-ə or /ˌwɪki-/ (listen) wik-ee-`) is completely useless for the model but will probably require many tokens. Any thoughts on that?

timoschick (approved):
Thanks a lot, @shanyas10 - looks good to me 👍

requirements.txt
```diff
@@ -4,3 +4,5 @@ wandb>=0.10.32,<1  # pip will likely update it to 0.12.1, but it is probably ok
 transformers>=4.6.0,<5  # pip will likely update it to 4.10.0, but it is probably ok and good for bugfixes.
 accelerate>=0.4.0,<1  # We may want to use 0.5.0 in the near future
 datasets[streaming]>=1.11.0,<2
+wikipedia2vec==1.0.5
+nltk==3.6.5
```
SaulLu, line 8: Really a tiny detail, but I think we can make these dependencies optional by adding them to the `setup.py` in `extras_require` :slightly_smiling_face: :
```
setup(
    name="bsmetadata",
    python_requires=">=3.7.11, <3.10",
    version="0.1.0",
    url="https://github.com/bigscience-workshop/metadata.git",
    author="Multiple Authors",
    author_email="xxx",
    description="Codebase for including metadata (e.g., URLs, timestamps, HTML tags) during language model pretraining.",
    packages=find_packages(),
    install_requires=install_requires,
    extras_require={
        "website_description_preprocessing": ["wikipedia2vec==1.0.5", "nltk==3.6.5"],
    },
)
```

bsmetadata/preprocessing_tools/website_desc_utils.py
SaulLu: Sounds like a great idea!

If it saves you some time @shanyas10 , here is a little code snippet that will remove the text in parentheses:
```python
import re

text = re.sub(r"\((?:[^)(]|\([^)(]*\))*\)", "", text)
```

On @timoschick's example, it will output: `Wikipedia  is a free content, multilingual online encyclopedia written and maintained by a community of volunteers through a model of open collaboration, using a wiki-based editing system.`

SaulLu (approved):
Thank you so much for all your hard work @shanyas10! I especially appreciate all the efforts you made to make this script run without internet access :confetti_ball: ! 

Thanks also for the tests!

bsmetadata/preprocessing_tools/website_desc_utils.py
shanyas10: Makes a lot of sense @timoschick . 
Thank you for the snippet @SaulLu . I've made the changes accordingly

requirements.txt
shanyas10: thank you for pointing this out @SaulLu :) I've made the required changes

shanyas10:
since the changes post approval are minor, I'm merging the PR :)

PR #53, opened 2021-10-29
feat: add html preprocessor and html parser

SaulLu:
Here is my PR adding:
- the HTML parser
- the HtmlPreprocessor
- some tests

This may need to be reworked depending on the format of the dataset that will arrive at this first preprocessor (cf comment) :smile: 

I am well aware that the HTML parser is not the easiest to digest, but if someone wants to take a look at it it would be with great pleasure!


bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
```diff
@@ -0,0 +1,459 @@
...
+        tags_exceptions_to_txt_max_min_chr_len_with_content: List[str] = None,
+    ):
+        self.html_str = html_str
+        self.tags_to_remove_with_content = tags_to_remove_with_content
+        self.tags_to_remove_alone = tags_to_remove_alone
+        self.attrs_to_keep = attrs_to_keep
+        self.start_parsing_at_tag = start_parsing_at_tag
+        self.convert_br_tag_to_breaking_line = convert_br_tag_to_breaking_line
+
+        self.tags_to_remove_alone = (
```
timoschick, line 260: Nitpick: If you'd change this to 
```python
if self.tags_to_remove_alone is None:
    self.tags_to_remove_alone = []
self.tags_to_remove_alone = [
    TagToRemove(...),
    ...
]
```
the list of fake tags doesn't need to be repeated.

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
```diff
@@ -0,0 +1,459 @@
...
+            tags_to_remove_with_content=tags_to_remove_with_content,
+        )
+
+    def apply(self):
+        html_str = self.html_str
+        # Traitement n°1: start the parsing at a special tags (mostly tested with <body>)
+        if self.start_parsing_at_tag is not None:
+            root = fromstring(html_str)
+            find = etree.XPath(f"//{self.start_parsing_at_tag}")
+            new_etree = find(root)[0]
```
timoschick, line 303: I guess this will throw an `IndexError` if the `self.start_parsing_at_tag` does not occur in the text (and I guess some html documents will not contain a `<body>` tag)? If so, it might make sense to wrap this error in some custom error class so that we can easily distinguish this "expected" error from unexpected ones.

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
```diff
@@ -0,0 +1,459 @@
...
+
+        # Traitement n°2: [all treatments impacting the chr_idx] we removes sub-trees from the HTML + we minify the html
+        html_str = htmlmin.minify(html_str, remove_comments=True, keep_pre=True)
+
+        new_etree = fromstring(html_str)
+
+        self._clean_etree(new_etree)
+
+        html_str = etree.tostring(new_etree, method="html", encoding="UTF-8", pretty_print=False).decode("UTF-8")
+        html_str = htmlmin.minify(html_str, keep_pre=True)
```
timoschick, line 319: I don't quite get why we need to call `minify` twice (but I guess it's not too important that I understand every little detail)

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
```diff
@@ -0,0 +1,459 @@
...
+
+    def _append_inline_element_separator(self, sb):
+        length = len(sb)
+        if length > 0:
+            last_buffer_char = sb[-1]
+            if last_buffer_char != PLAIN_TEXT_SEPARATOR and last_buffer_char != BLOCK_CONTENT_SEPARATOR:
+                sb += PLAIN_TEXT_SEPARATOR
+        return sb
+
+    def _get_text_and_metadata(self, root):
```
timoschick, line 408: The name is a bit confusing given that this function only returns the text. Maybe `_get_text_and_update_metadata`?

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
```diff
@@ -0,0 +1,459 @@
...
+            else {}
+        )
+        self.tags_to_remove_with_content = (
+            {tag_to_remove.tag: tag_to_remove for tag_to_remove in tags_to_remove_with_content}
+            if isinstance(tags_to_remove_with_content, list)
+            else {}
+        )
+
+        for tag_to_remove_characteristics in self.tags_to_remove_with_content.values():
+            if tag_to_remove_characteristics.method not in ["top-down", "bottom-up"]:
```
timoschick, line 79: Could you briefly explain the difference between top-down and bottom-up removal? I'm not sure I fully understand the difference from looking at the code.

timoschick (approved):
Unfortunately, I didn't have time to look at the code in great detail or check all edge cases, but form a high level, this looks really great! 🙌
I've added just a few comments about things I didn't fully understand and very minor suggestions.

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
SaulLu: That's a good suggestion, thanks! I just committed it :slightly_smiling_face:

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
SaulLu: This is indeed an interesting question! I added your proposal, it now raises a `ValueError`. Is this ok?

Details about the body tag :
From what I understand, the body tag is part of the mandatory tags for an html code (see [this link](https://www.understandingcode.com/html-basics/required-tags/) for example). It turns out that the method used above `fromstring` will try to make the passed string legitimate and if ever it turns out that the body tag is missing, it will add it (see [this example](https://colab.research.google.com/drive/1A2YSS4HlCxBbjda4G9p8_yMLw9oEF_-i?usp=sharing) in a google colaboratory).

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
SaulLu: This is indeed part of some of my choices as the function between the 2 minification operations `self._clean_etree` does not necessarily produce a minified html. 

It will probably be easier to see this on an example. Let's say that our initial html is :
```
\n    <html>\n    <head>\n    </head>\n    <body>\n    <h1>This is a title</h1>\n    <div>small</div>\n    <div>\n    <div>This is a <div>first</div> sub-div in div</div>\n    <div>This is a <div>second</div> sub-div in div</div>\n    </div>\n    <p>This is a paragraph not in div</p>\n    </body>\n    </html>\n
```
(I took the test example `test_parse_html_nested_example_min_length`). When we minify it the first time it becomes : 
```
<html><body> <h1>This is a title</h1> <div>small</div> <div> <div>This is a <div>first</div> sub-div in div</div> <div>This is a <div>second</div> sub-div in div</div> </div> <p>This is a paragraph not in div</p> </body> </html>'
```

But if now, we want to remove the subparts of the DOM corresponding to div tags if the contained text is longer than 7 characters, then the output of `self._clean_etree` will be : 
```
<html><body> <h1>This is a title</h1> <div>small</div>  <p>This is a paragraph not in div</p> </body> </html>
```

We notice in particular that there are two spaces between `</div>` and `<p>`. If we reapply minify, this double space will be reduced to 1:
```
<html><body> <h1>This is a title</h1> <div>small</div> <p>This is a paragraph not in div</p> </body> </html>
```

Please note, as we discussed a long time ago, the result of `minify` will not be a perfect minification but a "satisfactory" one.

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
SaulLu: Change committed :slightly_smiling_face:

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
SaulLu: image

Of course! Tags can be removed from the HTML DOM based on the length of the text contained in the tag.  

The "top-down" method will consist in going through the DOM from the root to the leaves and if during this traversal a node fulfills the conditions of elimination it will be deleted.

On the contrary, the "bottom-up" method will consist in going through the DOM from the leaves to the root and if during this traversal a node meets the conditions of elimination it will be deleted.

I implemented both but I think "bottom-up"  is a bit weird (by default "top-down"  is selected). I hesitated to delete  "bottom-up" but I thought that it was better not to lose the code if ever during the discussion this method appears to be useful.

bsmetadata/preprocessing_tools/html_parser/objects.py
```diff
@@ -9,6 +9,10 @@ class TagToRemove:
     content_max_char_length: float = float("inf")
 
 
+# The "top-down" method will consist in going through the DOM from the root to the leaves and if during this traversal
+# a node fulfills the conditions of elimination it will be deleted.
+# On the contrary, the "bottom-up" method will consist in going through the DOM from the leaves to the root and if 
+# during this traversal a node meets the conditions of elimination it will be deleted.
```
SaulLu, line 15: @timoschick, regarding your last comment, I've added a description here. :slightly_smiling_face:

bsmetadata/preprocessing_tools/html_parser/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    tags_to_remove_with_content: Optional[List[TagToRemoveWithContent]] = None,
+    tags_to_remove_alone: Optional[List[str]] = None,
+    attrs_to_keep: Optional[List[str]] = None,
+    consecutive_tags_to_fold: Optional[List[str]] = None,
+    convert_br_tag_to_breaking_line: Optional[bool] = False,
+    txt_max_chr_len_alone: float = -float("inf"),
+    txt_min_chr_len_alone: float = -float("inf"),
+    tags_exceptions_to_txt_max_min_chr_len_alone: List[str] = None,
+    txt_max_chr_len_with_content: float = -float("inf"),
+    txt_min_chr_len_with_content: float = -float("inf"),
```
cccntu, line 18: I don't fully understand the purpose of the variable, but shouldn't max be `inf` by default? (and min be `-inf`)

Also nit: I prefer`float("-inf")`  over `-float("inf")`

bsmetadata/preprocessing_utils.py
```diff
@@ -52,3 +54,40 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
     def _extract_timestamp_from_url(self, url: str) -> Optional:
         # This would have to be implemented.
         return None
+
+
+class HtmlPreprocessor(MetadataPreprocessor):
+    """todo"""
```
cccntu, line 60: I really more doc string here to understand the code. 😅

bsmetadata/preprocessing_tools/html_parser/__init__.py
SaulLu: All these `*max_chr_len*` and `*min_chr_len*` variables refer to conditions on the text to remove a tag or a tag and its content.

In this case, the tags whose content would be of length between `txt_min_chr_len_with_content` and `txt_max_chr_len_with_content`  would be removed from the html DOM .

So by setting both to `-float("inf")` the condition will never be verified and we will keep everything.

> Also nit: I preferfloat("-inf") over -float("inf")

I have no problem making this change but I would be curious to know why :slightly_smiling_face:

bsmetadata/preprocessing_utils.py
SaulLu: This is indeed an oversight. I have added it. I'm at your disposal if you need more.

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
SaulLu: I'm testing the code on tianjianjiang 's dataset and indeed several examples don't have a body tag and apparently the `fromstring` method doesn't add it. I'll investigate a bit more to see why this body tag is missing in some examples. 

If I need to make some changes for this new dataset, I will make them in another PR if it's ok for you.

bsmetadata/preprocessing_tools/html_parser/__init__.py
cccntu: That makes sense. Now I am curious why it's content of length within a range we want to remove, not outside of a range?

> > Also nit: I preferfloat("-inf") over -float("inf")

> I have no problem making this change but I would be curious to know why 🙂

Please Don't mind it. It's just I have never seen `-float("inf")` before, and my last brain cell warned me that it might be undefined. But now I think it's fine, it's just flipping one bit. No need to change it. 🙂

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
timoschick: Sure, sounds great! 👍

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
timoschick: Perfect, makes sense!

bsmetadata/preprocessing_tools/html_parser/__init__.py
SaulLu: It's true that what you propose makes sense, especially for this variable that would apply to all tags.

The reason why I arrived in this configuration is that I had started by introducing only the conditions of deletion of tag "case by case" in the list `tags_to_remove_with_content` and finally if we wanted to have the configuration that you propose it was "enough" to list twice the same tag but with conditions on the length of the content different (one for the lower bound and the other for the upper bound). I added that later on this condition for all the tags.

If we think that it is more useful to reverse the condition, maybe we can do it in a second PR, what do you think?

cccntu: > it was "enough" to list twice the same tag but with conditions on the length of the content different (one for the lower bound and the other for the upper bound)

That's a really clever design, I didn't realize we can list a tag more than once.

 > maybe we can do it in a second PR, what do you think? 🙂

I think we can merge this already, since the html test seems to work and this PR doesn't touch other part of the codebase. And I don't think we need to change that.

Thank you for working on this! 👍

cccntu (approved):
LGTM.

setup.py
```diff
@@ -9,6 +9,17 @@ def req_file(filename):
...
+    "html": ["lxml==4.6.3", "htmlmin==0.1.12"],
+    "entity": ["REL @ git+https://github.com/manandey/REL.git#egg=REL"],
+    "timestamp": ["bs_dateutil @ git+git://github.com/cccntu/dateutil@2.8.5"],
+    "website_description": ["wikipedia2vec==1.0.5", "nltk==3.6.5"],
+}
+
+preprocessing_dependencies = []
+for dependencies in preprocessing_requires.values():
+    preprocessing_dependencies.extend(dependencies)
+
```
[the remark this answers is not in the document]
cccntu: We can probably add `if available then import`, in the processors def file. But installing them all at once saves the trouble of seeing error then installing them one by one.

PR #55, opened 2021-11-02
Preprocess Entity Tags

manandey:
Hi @timoschick,

I have created this new PR after processing the comments from the previous PR #52. It would be great if you can have a look and let me know if it looks ready to be merged. I will close the previous PR then. Thanks!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "bsmetadata/vendor/dateutil"]
 	path = bsmetadata/vendor/dateutil
 	url = https://github.com/cccntu/dateutil
+[submodule "bsmetadata/vendor/REL"]
+	path = bsmetadata/vendor/REL
+	url = https://github.com/manandey/REL
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,4 @@
 [submodule "bsmetadata/vendor/dateutil"]
 	path = bsmetadata/vendor/dateutil
 	url = https://github.com/cccntu/dateutil
-[submodule "bsmetadata/vendor/REL"]
-	path = bsmetadata/vendor/REL
-	url = https://github.com/manandey/REL
+
--- a/README.md
+++ b/README.md
@@ -82,6 +82,15 @@ Below is a table showing ``relative_start_pos`` (`start`) and ``relative_end_pos
 | ``<i> ... </i><b></b>`` | 0 | 0 | 1 | 2 |
 | ``<i> ... <b></b></i>`` | 0 | 2 | 0 | 1 |
 
+## Pre-processing Metadata:
+
+### Entity Tags.
+
+**Pre-requisite steps for preprocessing Entity Tags**
+
+* Run `pip install -r bsmetadata/vendor/REL/requirements.txt`
+* Run `preprocessing_scripts/download_entity_processing_files.sh` to download all the required files.
+
 ## Contribute 🧠
 
 After installing the development dependencies, first you need to install the package in editable mode. You can do it by running in a bash at the root of the repository
--- a/README.md
+++ b/README.md
@@ -88,8 +88,7 @@ Below is a table showing ``relative_start_pos`` (`start`) and ``relative_end_pos
 
 **Pre-requisite steps for preprocessing Entity Tags**
 
-* Run `pip install -r bsmetadata/vendor/REL/requirements.txt`
-* Run `preprocessing_scripts/download_entity_processing_files.sh` to download all the required files.
+* Run `preprocessing_scripts/download_entity_processing_files.sh <"location of the folder to save the files">` to download all the required files. (approximate size of files to be downloaded in total is around 20GB and after extracting the zipped folders, space required is around 60GB.)
 
 ## Contribute 🧠
 
--- a/bsmetadata/preprocessing_scripts/download_entity_processing_files.sh
+++ b/bsmetadata/preprocessing_scripts/download_entity_processing_files.sh
@@ -1,0 +1,32 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+### This script downloads all the files required for the entity linking process in metadata preprocessing.
+### The Entity Linker used here is REL: Radboud Entity Linker [(Github URL:https://github.com/informagi/REL), (arxiv URL: https://arxiv.org/abs/2006.01969)]
+
+mkdir entity_linking_files
+cd entity_linking_files
+
+wget http://gem.cs.ru.nl/generic.tar.gz
+wget http://gem.cs.ru.nl/wiki_2019.tar.gz
+wget http://gem.cs.ru.nl/ed-wiki-2019.tar.gz
+
+tar -xzf generic.tar.gz
+tar -xzf wiki_2019.tar.gz
+tar -xzf ed-wiki-2019.tar.gz
+
+rm generic.tar.gz
+rm wiki_2019.tar.gz
+rm ed-wiki-2019.tar.gz
+
+cd ..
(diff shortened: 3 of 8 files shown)


timoschick (approved):
Sorry it took me so long to review the updated version 😓  I only looked at the changes briefly, but everything looks good to me. Still, I think it would be great if @SaulLu could also take a brief look.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "bsmetadata/vendor/dateutil"]
 	path = bsmetadata/vendor/dateutil
 	url = https://github.com/cccntu/dateutil
+[submodule "bsmetadata/vendor/REL"]
+	path = bsmetadata/vendor/REL
+	url = https://github.com/manandey/REL
```
SaulLu, line 6: Out of curiosity, why did you remove this dependency from `requirements.txt` and put it here? :blush: 

After discussing it offline with @cccntu , it seems that it is complicated to pip install `bsmetadata` with this `git submodule` dependency in a folder. How do you run your code currently? 

Personally, I would be in favor of adding an `extra_require` to the `bsmetadata` setup, which has the added advantage of making the dependency optional: we only install it if we want to do the preprocessing!

This will look like this in `setup.py`:
```python
setup(
    name="bsmetadata",
    ...
    extras_require={
        "entity_preprocessing": ["REL @ git+https://github.com/manandey/REL.git#egg=REL"],
    },
)
```
By changing well on the pointer to the branch, commit of your choice. See this [link](https://stackoverflow.com/questions/32688688/how-to-write-setup-py-to-include-a-git-repository-as-a-dependency/54794506#54794506) for the syntax to use

README.md
```diff
@@ -82,6 +82,15 @@ Below is a table showing ``relative_start_pos`` (`start`) and ``relative_end_pos
...
 | ``<i> ... <b></b></i>`` | 0 | 2 | 0 | 1 |
 
+## Pre-processing Metadata:
+
+### Entity Tags.
+
+**Pre-requisite steps for preprocessing Entity Tags**
+
+* Run `pip install -r bsmetadata/vendor/REL/requirements.txt`
+* Run `preprocessing_scripts/download_entity_processing_files.sh` to download all the required files.
```
SaulLu, line 92: I guess that currently the first step is to run the `git submodule update --init --recursive` command ? If so, it would be nice to add it to the readme (this addition will not be necessary if we go with the `extras_require` solution describe in the comment above :wink: )

bsmetadata/preprocessing_utils.py
```diff
@@ -80,3 +85,70 @@ def _extract_timestamp_from_url(self, url: str) -> Optional[str]:
         date = remove_improbable_date(date)
         date = str(date) if date is not None else None
         return date
+
+
+class EntityPreprocessor(MetadataPreprocessor):
+    """Metadata preprocessor for adding entity information."""
+
+    base_url = ".\preprocessing_scripts\entity_linking_files"
```
SaulLu, line 93: Would it be possible to set the path in the preprocessor `__init__` arguments? So we don't have to download the files (of several gigas) to this exact location. I think this will make it easier to use the installed package smile

bsmetadata/preprocessing_utils.py
```diff
@@ -80,3 +85,70 @@ def _extract_timestamp_from_url(self, url: str) -> Optional[str]:
...
+    wiki_version = "wiki_2019"
+
+    def __init__(self):
+        self.mention_detection = MentionDetection(self.base_url, self.wiki_version)
+        self.tagger_ner = load_flair_ner("ner-fast")
+        self.config = {
+            "mode": "eval",
+            "model_path": "ed-wiki-2019",
```
SaulLu, line 101: Is it a relative path to `self.base_url`?

README.md
```diff
@@ -82,6 +82,15 @@ Below is a table showing ``relative_start_pos`` (`start`) and ``relative_end_pos
...
+
```
SaulLu, line 93: It would also be perfect if we could indicate how much space it would take to make this download (this can be added in a future PR)

tests/test_entity_preprocessor.py
```diff
@@ -0,0 +1,45 @@
+import unittest
+
+from datasets import Dataset
+
+from bsmetadata.preprocessing_utils import EntityPreprocessor
+
+
+class TestEntityPreprocessor(unittest.TestCase):
```
SaulLu, line 8: Thank you very much for this test! It's really good to understand how the entity extraction works!

tests/test_entity_preprocessor.py
```diff
@@ -0,0 +1,45 @@
...
+    def test_extract_entities(self):
+
+        my_dict = {
+            "id": [0, 1, 2],
+            "text": [
+                "Paris is the beautiful place to visit",
+                "Obama was the guest of honor at the conference.",
+                "Bieber performed at the concert last night",
```
SaulLu, line 16: I would advise maybe to take an example where the entity is not at the beginning to check the `char_start_idx` and `char_end_idx.

SaulLu (commented):
Thanks a lot for all this work @manandey  (and for making something that works offline!) :muscle: 

My main comment concerns the use of the `bsmetadata` package as explained in my first comment.

manandey:
Thanks a lot for reviewing the PR @SaulLu! 🙂 I tried to address most of the suggestions. It would be great if you can do another round of review and let me know if any other changes need to be done from my side. Otherwise, feel free to merge the PR.

requirements.txt
```diff
@@ -4,3 +4,10 @@ wandb>=0.10.32,<1  # pip will likely update it to 0.12.1, but it is probably ok
 transformers>=4.6.0,<5  # pip will likely update it to 4.10.0, but it is probably ok and good for bugfixes.
```
manandey, line 4: The `torch` version used in the `requirements.txt` is `1.8.1` which seems to have some issue with the `flair` library I am using (https://github.com/flairNLP/flair/issues/2137). Can we use torch==1.9.0 instead?

SaulLu: We can definitely test this change (in another PR). The last version available on jz seems to be 1.9.0. :)

bsmetadata/preprocessing_scripts/download_entity_processing_files.sh
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+### This script downloads all the files required for the entity linking process in metadata preprocessing.
+### The Entity Linker used here is REL: Radboud Entity Linker [(Github URL:https://github.com/informagi/REL), (arxiv URL: https://arxiv.org/abs/2006.01969)]
+
+mkdir $1
+cd $1
```
SaulLu, line 18: Nice addition!

requirements.txt
```diff
@@ -4,3 +4,10 @@ wandb>=0.10.32,<1  # pip will likely update it to 0.12.1, but it is probably ok
...
 accelerate>=0.4.0,<1  # We may want to use 0.5.0 in the near future
 datasets[streaming]>=1.11.0,<2
+colorama
+konoha
+flair
+segtok
+Pillow>=7.1.0
+nltk
+anyascii
```
SaulLu, line 13: This seems to me to be similar to REL dependencies. Why do you need to add them here? :blush:

tests/test_entity_preprocessor.py
```diff
@@ -0,0 +1,50 @@
...
+            ],
+            "metadata": [[], [], []],
+        }  # toy dataset
+
+        target_id = [0, 1, 2]
+
+        target_text = [
+            "Paris is the beautiful place to visit",
+            "This Friday, Obama and Merkel will be meeting to discuss on issues related to climate change",
```
SaulLu, line 25: Nice!

bsmetadata/preprocessing_utils.py
```diff
@@ -80,3 +86,68 @@ def _extract_timestamp_from_url(self, url: str) -> Optional[str]:
...
+
+    def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
+
+        for example_text, example_metadata in zip(examples["text"], examples["metadata"]):
+            res = self._extract_entity_from_text(example_text)
+            result = self.postprocess_entity(res)
+            if not result:
+                continue
+            for val in result:
+                example_metadata.append(val)
```
SaulLu, line 113: Nit
```suggestion
            example_metadata.extend(result)
```

SaulLu (commented):
Thank you very much for making the changes that avoid using git submodules. I just have a small question about the new dependencies added to the `requirements.txt`

requirements.txt
manandey: Sorry, I fixed it in my latest commit. :) As discussed, I will be merging the PR then.

PR #56, opened 2021-11-05
Datasource and GenerationLength

chkla:
@timoschick my PR adds some new basic functions for processing the metadata types DATASOURCE and LENGTH:
- datasource parser (w/o threshold)
- datasource preprocessor
- datasource tests
- length parser (global)
- length preprocessor
- length tests


.gitignore
```diff
@@ -133,3 +133,4 @@ dmypy.json
 
 # PyCharm config files
 .idea/
+bsmetadata/vendor/dateutil
```
cccntu, line 136: What's this line for?  I don't think it's necessary.
I am not sure if this would cause a problem or not.

bsmetadata/preprocessing_utils.py
```diff
@@ -17,8 +17,16 @@
...
 
-from bsmetadata.vendor.dateutil.src.dateutil.parser import ParserError, parse
+import re
+from urllib.parse import urlparse
+import urllib
+
+import sys
+from bsmetadata import metadata_processors
+
+sys.path.insert(0, '/Users/christopher/git/metadata/')
```
cccntu: This line shouldn't be merged.

You can use it in your local copy, or you can try this in command line:
`export PYTHONPATH=$PYTHONPATH:$(pwd)`

cccntu (commented):
This is not a full review. I still need some time to understand the Submodule diff.

bsmetadata/preprocessing_utils.py
chkla: ✅

.gitignore
chkla: ✅ just my local workaround for working with submodules in visual code and not necessary for the public version

bsmetadata/preprocessing_utils.py
```diff
@@ -80,3 +83,86 @@ def _extract_timestamp_from_url(self, url: str) -> Optional[str]:
...
+        return examples
+
+    def _extract_length_from_text(self, text: str) -> Optional[str]:
+        return str(len(text))  # global
+
+
+class DatasourcePreprocessor(MetadataPreprocessor):
+    """An exemplary metadata preprocessor for adding datasource information based on URLs."""
+
+    def _check_numbers(self, sub_part):
```
timoschick, line 109: I think readability could be improved here by adding type hints (`_check_numbers(self, sub_part: List[str]) -> List[str]`) but I guess that's a matter of taste. At first, I thought that `sub_part` was a `str` and was confused why you would only remove the first/last character.

timoschick (commented):
Hi @chkla, thanks for your work 👍 This looks good to me overall, I've added a few comments.

bsmetadata/metadata_processors.py
```diff
@@ -178,6 +178,26 @@ def process_global(self, metadata_attrs: Dict[str, Any]) -> Optional[str]:
...
 
 
+class DatasourceProcessor(MetadataProcessor):
+    """An example metadata processor for datasource types."""
+
+    def process_global(self, metadata_attrs: Dict[str, Any]) -> Optional[str]:
+        # We represent the DATASOURCE by using meaningful information of the URL.
+        # URL: http://www.example.de/2015/forum/article/21-new-project
+        # Example: example.de forum article new project
+        return "".join(["Datasource", self.cfg.metadata_key_value_sep, metadata_attrs["value"]])
```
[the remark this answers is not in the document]
chkla: ✅ you're right, I've updated the example

bsmetadata/metadata_processors.py
```diff
@@ -178,6 +178,26 @@ def process_global(self, metadata_attrs: Dict[str, Any]) -> Optional[str]:
...
+
+
+class GenerationLengthProcessor(MetadataProcessor):
+    """An example metadata processor for the text length."""
+
+    def process_global(self, metadata_attrs: Dict[str, Any]) -> Optional[str]:
+        # We represent the length of a text by the number of characters.
+        # Example: Length: 123
+
+        return "".join(["Text Length", self.cfg.metadata_key_value_sep, metadata_attrs["value"]])
```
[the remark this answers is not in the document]
chkla: > And another quick thought: Did we decide to have generation length only at a global level? It might be useful to view this as a sentence-level (or `<p>` / `<div>` level, if we have HTML tags) kind of metadata.

✅ yeah, the local level was on my to-do list, and I've now added a new function to the length preprocessor to do this at the local (sentence-based) and global (text-based) level. but we can also think of a more complex splitting method than just using „dots“ to identify the relevant „parts" of a text.

bsmetadata/preprocessing_utils.py
chkla: ✅ good point, I like the idea. I've updated the function in this regard. thanks a lot

bsmetadata/metadata_processors.py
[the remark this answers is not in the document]
chkla: > Does the model always get the exact number? I'm not sure if we've discussed this already, but wouldn't it make sense to have some kind of bucketing (e.g., using powers of 2)?

🚧 If I remember correctly, we talked about alternative categories, e.g. small, medium, large, but not about a specific mapping between the number of characters and a category or so. So I think we decided to continue with the specific number of characters, but I would be also a fan of more abstract categories/ buckets as well.

manandey:
Merged in #81.

PR #65, opened 2021-11-22
Replace dateutil submodule with renamed package

### LINKED ISSUE #63: Rename dateutil submodule with another name
There are reports that the current approach might not work with `pip install`.
So I am planning to drop the submodule, and rename the modified dateutil so we can add it to requirements.

cc. @SaulLu @shanyas10

cccntu:
Fix #63

Please try this and let me know if it works.

@chkla You can try to merge this to your PR branch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "bsmetadata/vendor/dateutil"]
-	path = bsmetadata/vendor/dateutil
-	url = https://github.com/cccntu/dateutil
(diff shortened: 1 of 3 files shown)


SaulLu:
Thank you so much for working on this change! I see that the test failed when installing the packages? I am afraid that the format in the `extras_require` of `setup.py` is not exactly as expected.

Maybe you could try the format I had shared to manandey  for the entity tags? [link](https://github.com/bigscience-workshop/metadata/pull/55#discussion_r753380056)

cccntu:
@SaulLu  Thank you for the pointer! That saved me a lot of debugging.
The `#egg` suffix doesn't seem to be necessary so I dropped it. Is that ok? I don't really know what that means.
I also fixed a package related bug in `bs_dateutil` and bumped the version number.

SaulLu:
You're welcome! Everything is going well with the installation now :confetti_ball: !

SaulLu (approved):
You're welcome! Everything is going well with the installation now confetti_ball !

PR #72, opened 2021-11-25
feat: add checkpoints conversion and pushing to the hub script

SaulLu:
This PR proposes to add a script to convert all the checkpoints saved during the training to HF format + to create a branch per checkpoint on a repo and to push this checkpoint.

PS: I already used it to export 4 models on https://huggingface.co/bs-modeling-metadata

EDIT2: this script will need to be updated when #42  is merged

### DIFF
--- a/experiments/jz/utils/convert_checkpoint_to_hf_format.py
+++ b/experiments/jz/utils/convert_checkpoint_to_hf_format.py
@@ -1,0 +1,179 @@
+import dataclasses
+import gc
+import json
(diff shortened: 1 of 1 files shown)


SaulLu:
I merge this PR since these are scripts that have already been used for 4 experiments.

Nevertheless, feel free to have a look at it afterwards as it is never to late to change something and we'll have some new experiments that can benefit a change. :slightly_smiling_face:

PR #73, opened 2021-11-26
fix: change `bs_dateuil` github dependency in `setup.py`

SaulLu:
When I tried to install the package on JZ, I got an error (a timeout error if I remember correctly).

I'm not absolutely sure, but I think that the `git+git://github.com/cccntu/dateutil@2.8.5` dependency requires an ssh setup to perform the installation. As I don't have an ssh setup with my github account on jz, I suggest to change the dependency declaration for `bs_dateuil` as this package is hosted on a public repository to use an https connection. With this modification, the installation is successful on JZ. :slightly_smiling_face:


cccntu (approved):
LGTM

PR #75, opened 2021-11-29
add entity description while preprocessing entities

manandey:
This PR adds an additional field "ent_desc" having a single sentence description of the entities that are being extracted from the text.


bsmetadata/preprocessing_tools/entity_desc_utils.py
```diff
@@ -0,0 +1,29 @@
+import re
+import string
+from collections import defaultdict
+
+import nltk
+from wikipedia2vec.dump_db import DumpDB
+
+
+class EntityDescUtils:
```
shanyas10, line 9: Since the same code is also being used by Website Utils, I think we can merge them together to have a common utility class and use that instead

manandey: If there are no other issues apart from this, can we address this in a separate PR due to other priority tasks on my to-do?

SaulLu (commented):
Thanks a lot for this addition! 

But two small reflexions:
- I agree with @shanyas10, I might be great if we can put together the two utilities for websites descriptions and entities
- It's a good idea to add the description of entities. I wonder if we could not find a way (dataset / metric) to evaluate the disambiguation brought by this addition.

bsmetadata/preprocessing_tools/entity_desc_utils.py
manandey: Fixed!

manandey:
Thanks a lot for reviewing the code! :)

> * I agree with @shanyas10, I might be great if we can put together the two utilities for websites descriptions and entities
Fixed this.
> * It's a good idea to add the description of entities. I wonder if we could not find a way (dataset / metric) to evaluate the disambiguation brought by this addition.
Have pinged you over chat regarding this :)

bsmetadata/preprocessing_tools/wikipedia_desc_utils.py
```diff
@@ -40,3 +43,17 @@ def fetch_website_description_from_keyword(self, keyword: str) -> Optional:
...
+            text = self.wiki_dump_db.get_paragraphs(key)[0].text
+            text = re.sub(r"\((?:[^)(]|\([^)(]*\))*\)", "", text)
+            text = nltk.sent_tokenize(text)[0]
+        except KeyError:
+            try:
+                text = self.wiki_dump_db.get_paragraphs(self.redirects_map[keyword])[0].text
+                text = nltk.tokenize.sent_tokenize(text)[0]
+            except KeyError:
+                text = ""
+        return text
```
SaulLu, line 59: Can you detail a bit more which line is likely to return which error? 

I wonder if there is a way to de-interlace the exceptions. Or even to use the `fetch_wikipedia_description_for_title` method. 

Also, why in the second try/except, the `re.sub(r"\((?:[^)(]|\([^)(]*\))*\)", "", text)` processing is not necessary anymore?

If ever it is so interesting to have the second try/except I think the following formulation could bring clarity:

```suggestion
    def fetch_entity_description_from_keyword(self, keyword: str) -> str:
        key = string.capwords(keyword)
        text = self.fetch_wikipedia_description_for_title(key)

        if text is not None:
            key = self.redirects_map[keyword]
            text = self.fetch_wikipedia_description_for_title(key)
        
        if text is None:
            text = ""
        
        return text
```

SaulLu: @shanyas10 , I'd love to hear your opinion on this too. Wouldn't it make sense to call the "key" variables here "title"?

bsmetadata/preprocessing_utils.py
```diff
@@ -214,10 +218,17 @@ def postprocess_entity(self, resu_list):
                     resu_list[ent][0] + resu_list[ent][1]
                 ),  # element at index = 1 in the result list corresponds to length of the entity
                 "value": entity,
+                "ent_desc": ent_desc,
```
SaulLu, line 221: I would just like to draw @timoschick 's attention here as it concerns the final data format.

For the record, HTML metadata also has an additional key called [`html_attrs`](https://github.com/bigscience-workshop/metadata/blob/ecae719716384caa51d1cd564078ea0fb807da95/bsmetadata/preprocessing_tools/html_parser/objects.py#L53)

SaulLu (commented):
Thank you for taking our suggestion into account! :smile: 

I'm not quite sure I understand everything that is done in the `fetch_entity_description_from_keyword` method at the moment, I detailed that in the associated comment

bsmetadata/preprocessing_tools/wikipedia_desc_utils.py
shanyas10: Hi @SaulLu , yes indeed it would make more sense to call the "key" variables as title. Also having text = re.sub(r"\((?:[^)(]|\([^)(]*\))*\)", "", text) before returning the text would ensure that we don't have unnecessary tokens pulled from wikipedia descriptions.

bsmetadata/preprocessing_utils.py
```diff
@@ -171,7 +171,9 @@ def _extract_website_desc_from_url(self, url: str) -> Optional:
 class EntityPreprocessor(MetadataPreprocessor):
     """Metadata preprocessor for adding entity information."""
 
-    def __init__(self, base_url):
+    def __init__(self, base_url, path_wiki_db):
+        self.wiki_db_path = path_wiki_db
+        self.entity_utils = WikipediaDescUtils(path_wiki_db)
```
shanyas10, line 176: There's a small concern that I have here. Calling the init method of WikipediaDescUtils from two different processors would initialize the `dump_db `and `redirects_map` twice. While not being ideal design wise, this would also add up to the time since `redirects_map`, which takes some substantial time to be populated. would be created from scratch again. Instead, we can have it at a higher level (maybe at Metadata Processor level) or find another way, maybe cache or something which initializes this only once. cc: @SaulLu @timoschick

bsmetadata/preprocessing_tools/wikipedia_desc_utils.py
SaulLu: Thank you so much for sharing this info! 

@manandey , do you agree with renaming these variables?

manandey: @SaulLu, I have edited my previous comment. :)

manandey: @SaulLu, I have updated `key` to `title` as per your advice. :) Regarding using `fetch_wikipedia_description_for_title` method, I was not preferring to use it because in case it is unable to find any text for that `title/keyword`, it will return `None` and I prefer to have `""` instead in such a case.

bsmetadata/preprocessing_utils.py
shanyas10: Yes, absolutely. We can have a separate PR to handle this in the best manner :)

bsmetadata/preprocessing_tools/wikipedia_desc_utils.py
SaulLu: > fetch_wikipedia_description_for_title method, I was not preferring to use it because in case it is unable to find any text for that title/keyword, it will return None and I prefer to have "" instead in such a case.

Yes absolutely, I was just proposing to use the method inside yours :slightly_smiling_face:

PR #77, opened 2021-12-02
fix: reduce try/except scope into `fetch_wikipedia_description_for_title`

SaulLu:
As discussed offline with @shanyas10 , we need to have a try/except in the `fetch_wikipedia_description_for_title` method if the title is not in the database. 

On the other hand, we can be more precise on the type of error to be catched in order to have other types of errors raised (like for example when we forgot to install the punkt model necessary to`nltk.sent_tokenize`)


shanyas10 (approved):
Thank you so much for this change @SaulLu :)

PR #80, opened 2021-12-07
Update entity-tags preprocessing code to speed up the process

manandey:


bsmetadata/preprocessing_tools/wikipedia_desc_utils.py
```diff
@@ -45,16 +45,16 @@ def fetch_website_description_from_keyword(self, keyword: str) -> Optional:
...
+            title = string.capwords(keyword)
+            text = self.wiki_dump_db.get_paragraphs(title)[0].text
+            text = re.sub(r"\((?:[^)(]|\([^)(]*\))*\)", "", text)
+            text = nltk.sent_tokenize(text)[0]
+        except KeyError:
+            try:
+                text = self.wiki_dump_db.get_paragraphs(self.redirects_map[keyword])[0].text
+                text = nltk.tokenize.sent_tokenize(text)[0]
+            except KeyError:
+                text = ""
```
SaulLu, line 59: I have a small question, I was under the impression that we had discussed these lines of code in the previous RP, specifically [here](https://github.com/bigscience-workshop/metadata/pull/75#discussion_r761177998).

Could you give me a little more info on why you changed it?

manandey: Hi @SaulLu, when I tested the previous code that you had suggested, it was failing for certain cases where it was not able to find the description (throwing `KeyError`) using the code on the first `try` block. So, I changed it back so that instead of failing it can continue and try to use the `redirects_map` feature and see if it can still find the description using it. Otherwise, we set the value to `""`.

shanyas10: Few suggestions:
1. Have tokenization and removal of special characters logic outside try catch and pulled out commonly for both scenarios
2. Instead of having a nested try catch,  call fetch_description_from title, if it's none, check the redirects map -> if key exists, call fetch_desc_from_title else catch key error and return "" or None.

bsmetadata/preprocessing_tools/wikipedia_desc_utils.py
```diff
@@ -45,14 +45,13 @@ def fetch_website_description_from_keyword(self, keyword: str) -> Optional:
...
         title = string.capwords(keyword)
-        text = self.wiki_dump_db.get_paragraphs(title)[0].text
-        text = re.sub(r"\((?:[^)(]|\([^)(]*\))*\)", "", text)
-        text = nltk.sent_tokenize(text)[0]
+        text = self.fetch_wikipedia_description_for_title(title)
 
         if text is None:
-            text = self.wiki_dump_db.get_paragraphs(self.redirects_map[keyword])[0].text
-            text = nltk.tokenize.sent_tokenize(text)[0]
+            title = self.redirects_map[keyword]
```
shanyas10, line 53: Sorry if my previous comment was not very clear but I think this can throw a key error if the key is not found. We can wither handle it using try catch or keep a prior check ` if key in redirects_map:`

SaulLu: Oh yes, we should indeed tackle this issue if the keyword has a probability to not be in `redirects_map`.

I am in favor of your second suggestion @shanyas10 (the use of try/catch need to be avoided when we can IMO) :

```python
if text is None and keyword in self.redirects_map:
    title = self.redirects_map[keyword]
    text = self.fetch_wikipedia_description_for_title(title)
```

manandey: Updated.

manandey: Hope the main issue, ie. batching is fixed now. :)

SaulLu (approved):
Thank you very much for implementing the batching and making the "side" changes to the core of your PR. :slightly_smiling_face: :rocket:

bsmetadata/preprocessing_utils.py
```diff
@@ -182,62 +183,58 @@ def __init__(self, base_url, path_wiki_db):
             "mode": "eval",
             "model_path": "ed-wiki-2019",
         }
-        self.model = EntityDisambiguation(self.base_url, self.wiki_version, self.config)
+        self.model = EntityDisambiguation(self.base_url, self.wiki_version, self.config, reset_embeddings=True)
```
SaulLu, line 186: I remember that you said to me that this argument will speed the code, can you give more info about what it does ? :slightly_smiling_face:

manandey: According to this [PR](https://github.com/informagi/REL/issues/52), the performance improved a bit when `reset_embeddings` was set to `True`. 

Since the embeddings are stored in memory, the performance might become slower as the iterations progress and the memory becomes full. Setting `reset_embeddings` to `True`. resets the embeddings every time we run [get_data_items()](https://github.com/informagi/REL/blob/2bf75038729d54c79d57b8e7d5eac3d50c583690/REL/entity_disambiguation.py#L855) function in the `EntityDisambiguation` class and might improve the performance by a bit.

PR #81, opened 2021-12-08
update: generation length and datasource

chkla:
Hi @all, the last PR was now far behind the master branch, so I have updated the branch and it should be ready for the master branch. The PR includes:

* datasource parser (w/o threshold)
* datasource preprocessor
* datasource tests
* length parser (global and local)
* length preprocessor
* length tests

Best,
Chris


bsmetadata/preprocessing_utils.py
```diff
@@ -241,3 +244,137 @@ def fetch_mention_predictions(self, input_text: str) -> Optional:
         predictions, timing = self.model.predict(mentions_dataset)
         result = process_results(mentions_dataset, predictions, input_text)
         return result
+
+
+class GenerationLengthPreprocessor(MetadataPreprocessor):
+    """An exemplary metadata preprocessor for adding generation length information based on text."""
+
+    def __init__(self, mode):
```
timoschick, line 252: I would use "global" as the default value here

timoschick (approved):
I didn't carefully check the details, but from a high-level perspective, this looks good to me 👍

bsmetadata/preprocessing_utils.py
```diff
@@ -238,3 +240,137 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+        return examples
+
+    def _extract_length_from_text(self, text: str) -> Optional[str]:
+        """
+        Identify the length of a text.
+        """
+
+        return str(len(text))  # char-based length
+
+    def _extract_length_from_sentences(self, text: str) -> Optional[str]:
```
SaulLu, line 284: Reading the content of the method, I guessed that the output format is a `list` (and not a `str`). Are you agree?

bsmetadata/preprocessing_utils.py
```diff
@@ -238,3 +240,137 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+        meta_sentences = []
+
+        # Find all points in a text and store their absolute position to determine the final position of a sentence.
+        pos_sentences = [pos for pos, char in enumerate(text) if char == "."]
+
+        # Calculate the length of each sentence in a text based on a simple sentence splitting using the dots as indicators.
+        len_sentences = [self._extract_length_from_text(sent) for sent in text.split(".")]
+
+        # Iterate through the sentences of a text, storing the absolute beginning and end of each sentence and the associated length of each sentence.
+        for sent_pos, sent_len, i in zip(pos_sentences, len_sentences, range(len(len_sentences))):
```
SaulLu, line 298: here I have the impression that `sent_pos` and `sent_len` are not used. If this is the case, maybe we can simply a little bit this loop.

bsmetadata/preprocessing_utils.py
```diff
@@ -238,3 +240,137 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+        directories_parts = parts.path.strip("/").split("/")
+        directories_parts = self._clean_url_parts(directories_parts)
+
+        return parts.netloc + " > " + " > ".join(directories_parts)
+
+    def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
+
+        for example_url, example_meta in zip(examples["url"], examples["metadata"]):
+            example_datasource = self._extract_datasource_from_url(example_url)
+            print(example_datasource)
```
SaulLu, line 369: I think there was a little oversight here :slightly_smiling_face:

PR #103, opened 2021-12-15
Add code to sampling multiple metadata

cccntu:
According to discussion with @timoschick . 
Added a function to sample metadata as follows:
1. uniformly decide how many types of metadata should be included.
2. randomly choose metadata types to include.

WIP: need to add test and add code to actually call this function.


cccntu:
Updates: 
* Rebase and forced push
* Sample at least one, instead of 0
* Add (inverse) weighted sampling

Note: by default `data_config.metadata_config.random_sample_metadata` is set to `False`
We may want to set it to `True` in case we forget.

PR #104, opened 2021-12-20
change: retrieve the url from a column instead of from the metadata

SaulLu:
As the dataset we are building has a column with the url of each document (cc @tianjianjiang ), I suggest that we look for the url in this column rather than in the metadata column (this will be faster).

@shanyas10 and @cccntu  is that ok?

cc @timoschick for visibility


cccntu (approved):
LGTM

PR #106, opened 2021-12-21
add `path_or_url_flair_ner_model` in order to execute the entity extraction on a partition without internet

SaulLu:
To be able to run the code on a partition without internet, I need to download the flair model and to use it afterwards I need to fill in the path to the file in the `load_flair_ner` method.

cc @timoschick for visibility


manandey (approved):
LGTM! Thanks! :)

PR #111, opened 2021-12-21
Add joint training slurm script

cccntu:
The first commit is copy from template. So the changes from template can be seen from the diff.

### DIFF
--- a/experiments/jz/joint_training_toy/joint_training_toy1/02_load_dataset.slurm
+++ b/experiments/jz/joint_training_toy/joint_training_toy1/02_load_dataset.slurm
@@ -1,0 +1,62 @@
+#!/bin/bash
+#SBATCH --job-name=modelling-metadata-XX                                # (change me!) job name 
+#SBATCH --nodes=1
+#SBATCH --ntasks-per-node=1                                             # crucial - only 1 task per dist per node!
+#SBATCH --cpus-per-task=8                                               # (change me! between 0 and 48) number of cores per tasks 
+#SBATCH --hint=nomultithread                                            # we get physical cores not logical
+#SBATCH --gres=gpu:0                                                    # (change me! between 0 and 1) number of gpus 
+#SBATCH --time 00:10:00                                                 # (change me! between 0 and 20h) maximum execution time (HH:MM:SS) 
+#SBATCH --output=/gpfsdswork/projects/rech/six/uue59kq/logs/%x-%j.out   # output file name
+#SBATCH --error=/gpfsdswork/projects/rech/six/uue59kq/logs/%x-%j.err    # error file name
+#SBATCH --account=six@gpu                                               # account
+#SBATCH -p compil                                                       # partition with internet
+
+
+set -x -e
+
+# Next line will:
+# - load a conda environment with the dependencies on the master branch of github.com/bigscience-workshop/metadata/
+# - setup env vars ($HOME, $WORK, etc)
+# - load several modules (git)
+# Note: We can afford to have two conda environments: one stable for running experiments and one for development.
+# If there are new dependencies to install, you have to tell me about them and not do it in this script
+source $HOME/start-modelling-metadata-user
+
+# For the moment we can't directly use the new dataset feature on JZ which would avoid having to clone the dataset
+# repo from the HUB. So the first thing to do is to clone the repo of the XXX dataset if it does not already exist.
+HUB_REPO_NAME='XXX' # (change me! e.g. SaulLu/Natural_Questions_HTML_Toy_v2)
+
+# We define the name of the folder in which the clone will be made
+#Define multi-character delimiter
+delimiter="/"
+#Concatenate the delimiter with the main string
+string=$HUB_REPO_NAME$delimiter
+
+#Split the text based on the delimiter
+myarray=()
+while [[ $string ]]; do
+  myarray+=( "${string%%"$delimiter"*}" )
+  string=${string#*"$delimiter"}
+done
+REPO_DIR="${DATASETS_CUSTOM}/${myarray[-1]}"
+
+# We clone the repo if it doesn't exist
+if [[ -d "${REPO_DIR}" ]]
+then
+    echo "${REPO_DIR} already exists on your filesystem."
+else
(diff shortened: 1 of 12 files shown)


experiments/jz/joint_training_toy/joint_training_toy1/02_load_dataset.slurm
```diff
@@ -57,6 +57,7 @@ cd $WORK/repos/sync/metadata/
 
 # We check that the dataset can indeed be loaded
 python experiments/jz/utils/loading_script_utils/load_dataset.py \
-    dataset_name="${REPO_DIR}" \
-    train_file="XXX" \ # (change me and remove the comment! e.g "nq-train-*.jsonl.gz" or remove arg)
-    validation_file="XXX" # (change me and remove the comment! e.g. "nq-dev-*.jsonl.gz" or remove arg)
+    dataset_name="${REPO_DIR}" 
+    # \
+    # train_file="XXX" \ # (change me and remove the comment! e.g "nq-train-*.jsonl.gz" or remove arg)
+    # validation_file="XXX" # (change me and remove the comment! e.g. "nq-dev-*.jsonl.gz" or remove arg)
```
SaulLu, line 63: It seems to me that we still need to specify the files we want to include here (so that the automatic data format detection feature of the datasets library works)

For example by putting:
```
train_file="c4-train*"
```

experiments/jz/joint_training_toy/joint_training_toy1/03_create_dataset.slurm
```diff
@@ -44,14 +44,12 @@ REPO_DIR="${DATASETS_CUSTOM}/${myarray[-1]}"
 
 # Now we launch the script that will perform the preprocessing of the dataset
 # Feel free to add any arguments you like (change me!)
-python bsmetadata/train.py \ # (change me and remove the comment! if you have a specific script)
+python bsmetadata/train.py \
     data_config.experiment="with_metadata" \
-    data_config.metadata_config.metadata_list=["html"] \ # (change me and remove the comment!)
+    data_config.metadata_config.metadata_list=["entity", "timestamp", "url", "website_description"] \
```
SaulLu, line 49: Have you tested this command locally? I am not sure if this format is valid.

experiments/jz/joint_training_toy/joint_training_toy1/03_create_dataset.slurm
```diff
@@ -44,14 +44,12 @@ REPO_DIR="${DATASETS_CUSTOM}/${myarray[-1]}"
...
     data_config.metadata_config.max_seq_len=1024 \
     data_config.dataset_name="${REPO_DIR}" \
-    data_config.train_file="XXX" \ # (change me and remove the comment! e.g "nq-train-0\[0-2\].jsonl.gz" or remove arg)
```
SaulLu: Still not 100% sure, but I think that the automatic format detection feature need to know the file to include (like above)

experiments/jz/joint_training_toy/joint_training_toy1/04_do_training.slurm
```diff
@@ -1,10 +1,11 @@
 #!/bin/bash
-#SBATCH --job-name=modelling-metadata-html-XX                          # (change me!) job name
+#SBATCH --job-name=modelling-metadata-html-joint-toy-1
 #SBATCH --ntasks=1                    # number of MP tasks
 #SBATCH --constraint=v100-16g
 #SBATCH --cpus-per-task=8                                              # (change me! between 0 and 40) number of cores per tasks
 #SBATCH --hint=nomultithread                                           # we get physical cores not logical
 #SBATCH --time 01:00:00                                                # (change me! between 0 and 20h) maximum execution time (HH:MM:SS) 
+#SBATCH --gres=gpu:1                                                    # (change me! between 0 and 1) number of gpus
```
SaulLu, line 8: Just to be sure, we didn't want to test the parallel data training feature here? :relaxed:

SaulLu: I just read the comment on slack, it's fine for me if we test it later!

experiments/jz/joint_training_toy/joint_training_toy1/04_do_training.slurm
```diff
@@ -45,16 +46,12 @@ while [[ $string ]]; do
...
     data_config.metadata_config.max_seq_len=1024 \
     data_config.dataset_name="${REPO_DIR}" \
-    data_config.train_file="XXX" \ # (change me and remove the comment! e.g "nq-train-0\[0-2\].jsonl.gz" or remove arg)
-    data_config.validation_file"XXX" \ # (change me and remove the comment! e.g "nq-dev-00.jsonl.gz" or remove arg)
     data_config.preprocessing_num_workers=80 \
-    data_config.per_device_eval_batch_size=3 \
-    data_config.per_device_train_batch_size=3 \
     out_dir="${SCRATCH}/metadata_outputs/${SLURM_JOB_ID}" \
     jobid="${SLURM_JOB_ID}" \
     do_train=True \
```
SaulLu, line 57: As we didn't specify a file regex for the `validation_file`, I think we should put `do_eval=False` (or specify a file for the evaluation)

experiments/jz/joint_training_toy/joint_training_toy1/02_load_dataset.slurm
cccntu: I tried it and it seems to work the same with or without that argument.

experiments/jz/joint_training_toy/joint_training_toy1/04_do_training.slurm
cccntu: By default, it will do a 95-5 train-val split. I don't think it will be a problem. (I test ran it.)

experiments/jz/joint_training_toy/joint_training_toy1/03_create_dataset.slurm
cccntu: I updated it. It seems to work now. Thank you for reviewing.

PR #115, opened 2021-12-21
feat: change how the entity extraction process use ids

SaulLu:
Before this PR, the extraction of named entities required an "id" column, I propose with this PR to generate on the fly the ids needed for batching in the REL library. This feature allows to avoid having to design an ids column upstream.

cc @timoschick , @norakassner


manandey (approved):

manandey:
LGTM! Thanks! :)

PR #118, opened 2021-12-23
Add features types for the metadata to extract and test multiprocessing

SaulLu:
This PR add some variables that store the features type of the metadata extracted by each processor and the test are modified to test the multiprocessing


manandey (approved):

chkla (approved):

cccntu (commented):
What's the purpose of adding value type? Adding type-checking to the tests, I guess?

BTW, is it more memory efficient? I can't tell from the doc.

> features (Optional[datasets.Features], default None) – Use a specific Features to store the cache file instead of the automatically generated one.

SaulLu:
@cccntu 

> What's the purpose of adding value type? Adding type-checking to the tests, I guess?

The purpose is to use the multiprocessing feature (num_proc): without specifying the features type, we can end up with errors when the program reconciles the datasets calculated on the different processes :relaxed: 

> BTW, is it more memory efficient? I can't tell from the doc.

Unfortunately, I don't know either :slightly_smiling_face:

bsmetadata/utils.py
```diff
@@ -0,0 +1,20 @@
+import importlib.util
```
timoschick, line 1: I don't fully understand why we need all of this logic. Wouldn't it make more sense to just directly add `datasets` to the project requirements and replace

```python
if _datasets_available:
    from datasets import Value
```
with `from datasets import Value`?

manandey (approved):

manandey:
LGTM, thanks @SaulLu!

timoschick (commented):
LGTM

bsmetadata/utils.py
SaulLu: I just thought that if we wanted to use this part without datasets it wouldn't block (because in the end datasets is not the only lib available). But if it confuses you, I can remove it. :blush:

SaulLu: `datasets` is already in the `requirements.txt`

SaulLu: @timoschick , as for me this feature is not at all as important as the rest, I removed it in my last commit :slightly_smiling_face:

PR #119, opened 2021-12-24
create dataset with html, timestamp, url, datasource, generation length and website description metadata and tittles, footers and headers from HTML

SaulLu:
This PR includes:
- the python script to extract all the metadata from our initial dataset
- slurm scripts used to explore the hyperparameters to choose for extraction
- modifications on the HTML processor to 1) avoid error return and 2) extract data in new columns (head and footers and titles)
- An error wrapper to prevent a single example from crashing an entire job

What is missing in this PR:
- SLURMS scripts to extract the website descriptions and entities on the full dataset


bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
```diff
@@ -291,25 +300,48 @@ def __init__(
...
+                # this is not a valid HTML (begin probably with <?xml version="1.0")
+                logger.warning(f"This example wasn't parsed: invalid HTML")
+                return "", [], {}
+            for tag in self.tags_sub_tree_to_isolate:
+                find = etree.XPath(f"//{tag}")
+                matches = find(root)
+                html_extracted = [
+                    etree.tostring(match, encoding="UTF-8", pretty_print=False).decode("UTF-8") for match in matches
+                ]
+                additional_columns[tag] = html_extracted
```
SaulLu, line 318: This addition allows you to keep the head section and footer sections of the html file in other columns that we parse

bsmetadata/preprocessing_tools/html_parser/filters_and_cleaners.py
```diff
@@ -291,25 +300,48 @@ def __init__(
...
-                raise ValueError(
+                root = fromstring(html_str)
+            except ValueError:
+                # this is not a valid HTML (begin probably with <?xml version="1.0")
+                logger.warning(f"This example wasn't parsed: invalid HTML")
+                return "", [], {}
+            find = etree.XPath(f"//{self.start_parsing_at_tag}")
+            matches = find(root)
+            if len(matches) == 0:
+                logger.warning(
```
SaulLu, line 331: These additions avoid two types of errors encountered:
- a file that the `lxml` library cannot parse as html
- the case where the html does not contain a "body" tag

bsmetadata/preprocessing_tools/html_parser/objects.py
```diff
@@ -51,8 +51,8 @@ def convert_html_metadata_dataclass_to_dict(metadata: Metadata):
             "relative_end_pos": metadata.relative_end_pos,
             # The information about the HTML tag is separated into two keys because the dictionary must have a stable
             # format between the different types of metadata
-            # the comment have a specific type that is not recognized by pyArrow when we load it in a dataset
-            "value": str(metadata.value.tag) if metadata.value.tag == Comment else metadata.value.tag,
+            # some tags have a value of type `cython_function_or_method` which is not supported by pyrarrow
+            "value": str(metadata.value.tag) if type(metadata.value.tag) == type(Comment) else metadata.value.tag,
```
SaulLu, line 55: This addition avoids another type of error when working with the datasets library (it does not accept cpython formats)

bsmetadata/preprocessing_utils.py
```diff
@@ -591,3 +622,86 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
 
         examples[self.col_to_store_metadata] = example_metadata_list
         return examples
+
+
+class ErrorWrapperPreprocessor:
```
SaulLu, line 627: This class is useful for mass extraction: if I have an example that returns an error it does not prevent the processing of other examples and it allows me to "know" the type of error returned on this particular example (what we do not see otherwise)

experiments/jz/dataset/c4/README.md
```diff
@@ -0,0 +1,38 @@
+1. Download `wiki_en_dump.db`
```
SaulLu, line 1: This file contains instructions for downloading everything you need from jz

PR #123, opened 2022-01-14
Add fp16, multi-GPU training script (toy dataset)

cccntu:
* Changed an argument in load_dataset so it tries to read private datasets.
* Added gradient_step logging, so we can see time v.s. gradient step on wandb

* Adds 2 sub-experiments
  1.  + fp16
  2.  + fp16, 2 GPU
     * note: I only have 1 GPU, so I couldn't test this one, but I included a `local_test.sh` that can be used to test.

* The toy dataset used is now privated. To run the experiments, you'd need to configure the huggingface access tokens with `huggingface-cli  login`.

### DIFF
--- a/bsmetadata/experiments/with_metadata.py
+++ b/bsmetadata/experiments/with_metadata.py
@@ -11,6 +11,9 @@ from bsmetadata.metadata_utils import add_metadata_and_chunk_examples
 logger = logging.getLogger(__name__)
 
 
+load_dataset = functools.partial(load_dataset, use_auth_token=True)
+
+
 def get_dataloaders(tokenizer, args):
     """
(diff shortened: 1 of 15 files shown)


bsmetadata/experiments/with_metadata.py
```diff
@@ -11,6 +11,9 @@
 logger = logging.getLogger(__name__)
 
 
+load_dataset = functools.partial(load_dataset, use_auth_token=True)
```
SaulLu, line 14: This is something we don't necessarily need in our workflow I think as we work on the dataset already cloned locally. After that, I understand that you may need it for other tests on your side. :relaxed:

PR #132, opened 2022-01-22
build: pin datasets to 1.17.0

tianjianjiang:
For errors like https://github.com/bigscience-workshop/metadata/runs/4900291708?check_suite_focus=true#step:5:355

Technically I think it is not directly about datasets but pyarrow or similar. But since pyarrow's version does not change and we only control the version of datasets...


manandey (commented):
Thanks!

PR #134, opened 2022-01-27
`Title` preprocessor

manandey:
This PR has the implementation and the tests for the `TitlePreprocesor`

cc. @SaulLu


SaulLu:
Hi Manan,

As pointed out earlier, I unfortunately think that a piece is still missing. Currently the titles values aren't cleaned. 

You can find examples on the dataset explorer [here](https://huggingface.co/spaces/bs-modeling-metadata/dataset-exploration). In particular, the value for the first example of our dataset is:
```
"html_title":[
"<title>My Thoughts On It » Dad, I want to be an inventor</title>&#13; "
]
```

I think that we need to remove `<title>` and  `</title>&#13; ` but we'll have to look at more example to be sure that there isn't more noise to remove! :slightly_smiling_face:

timoschick (approved):

bsmetadata/preprocessing_utils.py
```diff
@@ -618,6 +618,49 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+            if self.col_to_store_metadata in examples
+            else [[] for _ in range(len(examples[self.col_text]))]
+        )
+
+        # Iterate through the metadata associated with all examples in this batch.
+        for example_title, example_metadata in zip(examples[self.col_title], example_metadata_list):
+
+            if not example_title:
+                continue
+            title = example_title[0]
```
SaulLu, line 654: Just to leave a trace
```suggestion
            # The number of titles retrieved on a page is not necessarily equal to 1. Here the choice is made to keep only the first title retrieved when there is one.
            if not example_title:
                continue
            title = example_title[0]
```

bsmetadata/preprocessing_utils.py
```diff
@@ -618,6 +618,49 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List]:
...
+            title = re.search("<title>(.*)</title>", title)
+            title = title.group(1)
```
SaulLu, line 656: If there is no match, `re.search("<title>(.*)</title>", title)` will output a `None` that doesn't have a `group` method. I think that it might be safer to handle this case.

SaulLu (commented):
Thanks for this addition! I've just left a few comments

bsmetadata/preprocessing_utils.py
manandey: Hi @SaulLu, I have updated the code to handle this case. Feel free to let me know if any other changes are required. Thanks! 🙂

SaulLu (approved):
Thanks for the last changes! Looks good to me! :+1:

PR #136, opened 2022-02-17
Fix accelerate not using multi-GPU 

cccntu:
Changed one line. 
It now works on my machine with 2 GPU.

c.c. @norakassner


SaulLu:
Should we test this fix or are we sure that it will work ? :slightly_smiling_face:

PR #137, opened 2022-02-19
Filter examples by `num_chars` to include in a batch

manandey:
Hi @SaulLu, 

I have updated the `EntityPreprocessor` to include an example in a batch only if its length is less than or equal to `num_chars`. I have set the default value to 5000 characters. Feel free to set a different value. Thanks!

### DIFF
--- a/bsmetadata/preprocessing_utils.py
+++ b/bsmetadata/preprocessing_utils.py
@@ -302,7 +302,7 @@ class EntityPreprocessor(
     def __init__(
         self,
         base_url,
-        num_chars,
+        num_chars=4000, #default length of text set as threshold to include an example in a batch
         path_or_url_flair_ner_model="ner-fast",
         col_to_store_metadata="metadata",
         col_text="text",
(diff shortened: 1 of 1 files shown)


timoschick:
Thanks for implementing this so quickly @manandey 👍  Could you quickly comment on how you chose the 5000 tokens? Or, more importantly: Do you have an estimate on how much data we loose if we set that threshold, i.e., how many documents will not get annotated?

manandey:
Hi @timoschick, I selected 5000 characters as some random value to start with. :) The reason behind selecting it in the first place was, when I ran these two [experiments](https://wandb.ai/manandey/entity%20extract%20new?workspace=user-manandey), a standard value for the (number of characters)/ (batch = 100 examples) was in the range (5000-6000). But we can definitely increase it to a larger value I guess since we might be using twice the RAM we were using when we faced the OOM error.

SaulLu:
We didn't use this in the end no?

manandey:
> We didn't use this in the end no?

@SaulLu, yeah, we didn't use it at the end. Maybe Mike merged it by mistake. Should I revert it?

manandey:
@tianjianjiang, no worries at all! @SaulLu, reverted the changes! :)

PR #139, opened 2022-02-28
build: sync setup.py defined dependencies and fix broken ones

tianjianjiang:
Mostly for obstacles I encountered for collecting metadata.
1. Our custom `Features` dict needs datasets 1.18.3
2. wikipedia2vec doesn't support Python 3.9
3. gensim 4 breaks wikipedia2vec
4. torch 1.8 breaks flair

----
TODO:
- [x] ~~datasets 1.18.3 seems breaking some tests.~~ It was pyarrow>=7 being rigorous about typing.


SaulLu (commented):
Thanks Mike! 

Indeed our dataset in the `jsonl.gz`format needs really new versions to be loadable. Another version constraint that we need is `pyarrow>=6.0.0`. Do you think that you can include this upgrade constraint as well? :slightly_smiling_face:

tianjianjiang:
> Thanks Mike!
> 
> Indeed our dataset in the `jsonl.gz`format needs really new versions to be loadable. Another version constraint that we need is `pyarrow>=6.0.0`. Do you think that you can include this upgrade constraint as well? 🙂

@SaulLu Sorry for letting this fall through the cracks. For `pyarrow` I thought `datasets` had that but I was mistaken. I will make it happen with our constrants.

PR #144, opened 2022-04-07
Entity at paragraph level

manandey:
close: #149

Preprocessor for entities at paragraph level. Tests will be updated soon.


tianjianjiang (approved):
LGTM!

bsmetadata/preprocessing_utils.py
```diff
@@ -716,6 +716,103 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List[OneToOneFeatur
...
+                # Initialize the start and end index of an entity
+                start_index = entity["char_start_idx"]
+                end_index = entity["char_end_idx"]
+                check = True
+
+                # Search the start and end index of paragraph in between which the entity is present without any duplicate entity values.
+                for paragraph in example_paragraph:
+                    # If paragraph["key"] != "paragraph", we skip this paragraph. Added this check since tests were failing.
+                    if paragraph["key"] != "paragraph":
+                        check is False
```
tianjianjiang, line 775:
```suggestion
                        check = False
```
I guess this is the indented assignment?

bsmetadata/preprocessing_utils.py
```diff
@@ -716,6 +716,103 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List[OneToOneFeatur
...
+                        continue
+                    if start_index >= paragraph["char_start_idx"] and end_index <= paragraph["char_end_idx"]:
+                        # Update the start and end index of an entity
+                        start_index = paragraph["char_start_idx"]
+                        end_index = paragraph["char_end_idx"]
+                        break
```
tianjianjiang, line 781:
```suggestion
                        check = True
                        break
```
And then another assignment is probably required?

bsmetadata/preprocessing_utils.py
```diff
@@ -716,6 +716,103 @@ def preprocess(self, examples: Dict[str, List]) -> Dict[str, List[OneToOneFeatur
...
+
+            # Iterate through the entities associated with this example.
+            for entity in example_entity:
+                # If entity["key"] != "entity", we skip this entity. Added this check since tests were failing.
+                if entity["key"] != "entity":
+                    continue
+                # Initialize the start and end index of an entity
+                start_index = entity["char_start_idx"]
+                end_index = entity["char_end_idx"]
+                check = True
```
tianjianjiang, line 769:
```suggestion
                start_index = None
                end_index = None
```
I guess this will suffice for the initialization.

tests/test_preprocessing_utils.py
```diff
@@ -536,6 +537,41 @@ def create_data(self):
...
+                {
+                    "char_end_idx": 120,
+                    "char_start_idx": 0,
+                    "key": "entity_paragraph",
+                    "relative_end_pos": 1,
+                    "relative_start_pos": 1,
+                    "type": "local",
+                    "value": "Angela_Merkel",
+                },
+            ],
```
tianjianjiang, line 561: Just want to be sure I understand the intention: the two examples mean "a paragraph [0,120) containing two entities; the first one is Barack_Obama and the second is Angela_Merkel." Is it correct?

bsmetadata/preprocessing_utils.py
manandey: I think we require this, since we need to initialize it to the `char_start_idx` and `char_end_idx` of the entity to later compare it to the `char_start_idx` and `char_end_idx` of the paragraph in line 777.

bsmetadata/preprocessing_utils.py
manandey: Code quality test was failing because of this.

bsmetadata/preprocessing_utils.py
manandey: This won't be required if we set it to `True` in the outer loop itself, in line: 769.

tests/test_preprocessing_utils.py
manandey: Yes, you are correct!

bsmetadata/preprocessing_utils.py
tianjianjiang: @manandey For https://github.com/bigscience-workshop/metadata/runs/6047523378?check_suite_focus=true#step:5:13 , I guess it was `check == False` or `if check == False` and yet it seemed an assignment rather than a comparison. When `check is False` without an `if`, `check` isn't changed.

bsmetadata/preprocessing_utils.py
tianjianjiang: > This won't be required if we set it to `True` in the outer loop itself, in line: 769.

@manandey Indeed. I was only making a strict assumption that the inner loop of paragraph may never be true for `start_index >= paragraph["char_start_idx"] and end_index <= paragraph["char_end_idx"]`, but of course it is unreal for our current dataset. So feel free to ignore this one and the related one below.

bsmetadata/preprocessing_utils.py
tianjianjiang: Same as https://github.com/bigscience-workshop/metadata/pull/144#discussion_r851809361

PR #147, opened 2022-04-15
Pos processing website desc

shanyas10:


manandey (approved):

tianjianjiang:
@shanyas10 Thank you for the PR! Only one small thing about the integration test (not in this PR though): the "same column" test requires a double-check of metadata key. In the interest of time-zone, it may be easier for you to get more details from @manandey since he has encountered the situation before.

PR #155, opened 2022-06-30
Finalizing for big training

cccntu:
# What is in this PR

## Important Stuff
- Add a validation set without metadata, and filter out examples without metadata in each metadata evaluation set.
- Always pad examples to max length for no-metadata validation set, otherwise deepspeed errors in validation.
- Fix a preprocessing error that happens when there is no metadata by filtering out examples with empty "text" column (there isn't that many of them).
- Fix some logic that caused some metadata not being processed
  - Fix v2 to v1 conversion, so it doesn't overwriting metadata key.
  - Add `metadata_column_list`, separating from `metadata_list` for v2 data format
  - Fix html preprocessor class attribute name. @SaulLu 

## Other Stuff
- Add an option to limit the size of the evaluation set.
- Add an option to evaluate/save at specific steps (to test it early durning training).
- Refactor V2 dataset loading and preprocessing code so it is easier to work with.
- Add a config file for v2

## TODO

- Do the TODOs in the code
- Update documentation
- Add dataset streaming mode
   * validation set don't need that
   * does it work when we specify the data files?
   * use https://huggingface.co/docs/datasets/v2.3.2/en/package_reference/main_classes#datasets.interleave_datasets with `probabilities`, to mix the datasets with/without entity based on the number of files in each set.

### DIFF
--- a/bsmetadata/input_pipeline.py
+++ b/bsmetadata/input_pipeline.py
@@ -55,9 +55,7 @@ class DataConfig:
     )
     validation_size_max: Optional[int] = field(
         default=None,
-        metadata={
-            "help": "The maximum number of samples to use for validation. If None, all samples will be used."
-        },
+        metadata={"help": "The maximum number of samples to use for validation. If None, all samples will be used."},
     )
 
 
--- a/bsmetadata/input_pipeline.py
+++ b/bsmetadata/input_pipeline.py
@@ -53,6 +53,12 @@ class DataConfig:
             " the dataset. If you are using `with_metadata` the recommended batch size is 1.."
         },
     )
+    validation_size_max: Optional[int] = field(
+        default=None,
+        metadata={
+            "help": "The maximum number of samples to use for validation. If None, all samples will be used."
+        },
+    )
 
 
 def get_dataloaders(tokenizer, cfg: DataConfig):
--- a/bsmetadata/input_pipeline.py
+++ b/bsmetadata/input_pipeline.py
@@ -57,6 +57,10 @@ class DataConfig:
         default=None,
         metadata={"help": "The maximum number of samples to use for validation. If None, all samples will be used."},
     )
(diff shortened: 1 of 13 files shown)


bsmetadata/metadata_processors.py
```diff
@@ -266,7 +270,9 @@ def __init__(
         txt_min_chr_len = cfg.html_parser_config.all_tags_rules.txt_min_chr_len
         tags_exceptions = cfg.html_parser_config.all_tags_rules.tags_exceptions_to_txt_max_min_chr_len
         tags_to_remove_alone = [
-            html_parser.objects.TagToRemove(tag=tag, txt_max_chr_len=txt_max_chr_len, txt_min_chr_len=txt_min_chr_len)
+            html_parser.objects.TagToRemove(
```
SaulLu, line 273: That sounds good! I don't know when and how this anomaly crept in...

cccntu:
Hi @tianjianjiang, I think that's some permission issue, you should `sudo rm`  this existing folder, or change the destination, or comment out the code that saves validation if you don't need that.

manandey (approved):

bsmetadata/input_pipeline.py
```diff
@@ -53,6 +53,14 @@ class DataConfig:
             " the dataset. If you are using `with_metadata` the recommended batch size is 1.."
         },
     )
+    validation_size_max: Optional[int] = field(
+        default=None,
+        metadata={"help": "The maximum number of samples to use for validation. If None, all samples will be used."},
+    )
+    streaming: bool = field(
+        default=False,
+        metadata={"help": "Wether to use streaming mode. Note that validation will not be streamed."},
```
manandey, line 62:
```suggestion
        metadata={"help": "Whether to use streaming mode. Note that validation will not be streamed."},
```

PR #156, opened 2022-07-01
new configuration for HTML tags

SaulLu:
This PR proposes to change the configurations for the HTML tags and also for the entities (cc @manandey )

### DIFF
--- a/bsmetadata/hydra_configs/v2.yaml
+++ b/bsmetadata/hydra_configs/v2.yaml
@@ -39,7 +39,14 @@ data_config:
         - id
         txt_max_chr_len: 0
         txt_min_chr_len: -.inf
-        tags_exceptions_to_txt_max_min_chr_len: null
+        tags_exceptions_to_txt_max_min_chr_len:
+        - tabletr
+        - th
+        - td
+        - colgroup
+        - thead
+        - tfoot
+        - tbody
       tags_to_remove_alone_tag_name:
       - body
       tags_to_remove_alone_txt_max_chr_len:
@@ -54,16 +61,10 @@ data_config:
   train_file: c4-en-html_cc-main-2019-18_pq00-001.jsonl.gz
   validation_file: c4-en-html_cc-main-2019-18_pq00-000.jsonl.gz
   overwrite_cache: false
-  cache_dir: null
-  extension: null
   preprocessing_num_workers: 6
   validation_split_percentage: 5
-  block_size: null
-  map_batch_size: 1
 weight_decay: 0.01
 learning_rate: 5.0e-05
-gradient_accumulation_steps: 8
-num_train_epochs: 1
 max_train_steps: 100
 lr_scheduler_type: linear
 num_warmup_steps: 6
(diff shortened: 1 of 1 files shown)


bsmetadata/hydra_configs/v2.yaml
```diff
@@ -27,8 +27,8 @@ data_config:
     metadata_sep: ' | '
     metadata_key_value_sep: ': '
     metadata_probability: 1.0
-    treat_local_metadata_as_regular_text: false
-    add_local_metadata_special_tokens_in_prefix: false
+    treat_local_metadata_as_regular_text: true
```
SaulLu, line 30: If I recall correctly this argument is used to know if we want to take the local metadata occurrences into account in the loss: I think that we want them to be taken into account.

Do you agree @manandey , @cccntu ?

bsmetadata/hydra_configs/v2.yaml
```diff
@@ -27,8 +27,8 @@ data_config:
...
+    add_local_metadata_special_tokens_in_prefix: true
```
SaulLu, line 31: I would suggest adding a parameter for the local metadata as well to allow their controllability afterwards.
A prompt beginning with `html |||` would triggers the addition of html tags inside afterwards. If the prompt doesn't begin with `html |||`, then we shouldn't have html tags in the text generated

bsmetadata/hydra_configs/v2.yaml
```diff
@@ -37,33 +37,15 @@ data_config:
...
-        - caption
-        - colgroup
-        - thead
-        - tfoot
-        - tbody
-        - li
-        - ol
-        - ul
-        - span
+        tags_exceptions_to_txt_max_min_chr_len: null
```
SaulLu, line 42: I'm not entirely sure about the yaml syntax here, but the idea is that I don't want to define exceptions (so in the code this should result in an empty list I believe)

bsmetadata/hydra_configs/v2.yaml
manandey: Yeah, I completely agree! :)

manandey (approved):
Thanks, Lucile!

bsmetadata/hydra_configs/v2.yaml
cccntu: Yes, I think it makes more sense to calculate the loss of html, too.

bsmetadata/hydra_configs/v2.yaml
cccntu: > A prompt beginning with html ||| would triggers the addition of html tags inside afterwards. If the prompt doesn't begin with html |||, then we shouldn't have html tags in the text generated

Sound good to me!

> I would suggest adding a parameter for the local metadata as well to allow their controllability afterwards.

I don't understand this. Does this change already does what you want? Or are you suggesting to add a different parameter?

SaulLu: > Does this change already does what you want? 

yes exactly! My comment applies to all local-type metadata (for a long time the entities were also local-type metadata, I don't know what is the status currently). The idea is exactly what I explained in the following sentences: `html` is the parameter for the local metadata "html" :smile: 

Don't hesitate to tell if this is still confusing :smile:

manandey: > for a long time the entities were also local-type metadata, I don't know what is the status currently

Entities are still local-type metadata. :slightly_smiling_face:

PR #157, opened 2022-07-04
add new configs for entity_paragraph

manandey:
Hi @SaulLu, I have added the configs for entity_paragraph in this PR and made a few other minor updates.

### DIFF
--- a/bsmetadata/hydra_configs/v2.yaml
+++ b/bsmetadata/hydra_configs/v2.yaml
@@ -4,7 +4,6 @@ data_config:
     random_sample_metadata: false
     metadata_list:
     - html
-    - entity
     - timestamp
     - website_description
     - title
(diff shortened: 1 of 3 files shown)


manandey (commented):
@cccntu, in this config I have both entity and entity_paragraph, but incase if we decide on removing entity from here at the end, then I would like to replace this [line](https://github.com/bigscience-workshop/metadata/blob/4f857671ebc0eb6e6a85434d86fb9c8b44f71a2b/bsmetadata/hydra_configs/v2.yaml#L26)  with 
`local_metadata_special_tokens:
      entity_paragraph: "entity"`

cccntu:
Hi @manandey ,
Thanks for the PR.
I think we can leave this open until we decide what config we want.

cccntu (approved):
LGTM

PR #160, opened 2022-08-14
feat: log the average of the loss rather than the value of the main process

SaulLu:
As discussed during the meeting this week, here is the change needed to average the loss logged on WANDB on all GPUs.

I tested the change and observed a modification of the logged values:
image


cccntu (approved):
LGTM, thanks!

PR #174, opened 2022-12-16
Fix code quality tests

manandey:


bsmetadata/train.py
```diff
@@ -297,7 +297,10 @@ def main(args: CFG) -> None:
             model, optimizer, dummy_dataloader, scheduler
         )
     else:
-        format_fn = lambda x: x
```
manandey, line 300: @cccntu just for your reference, made the following changes to fix the code quality changes. Hope it's fine.

PR #188, opened 2023-04-28
Update metadata_utils.py

jordiclive:
fairer without metadata. If you keep the metadata_probability the same, the idea is to give the same context in each example in a without_metadata run.

### DIFF
--- a/bsmetadata/metadata_utils.py
+++ b/bsmetadata/metadata_utils.py
@@ -128,7 +128,6 @@ def add_metadata_and_chunk_examples(
                 total_len = len(text_chunk_encoded)
                 padding_len = cfg.max_seq_len - len(text_chunk_encoded)
                 attention_mask = [1] * total_len + [0] * padding_len
-                text_chunk_encoded = text_chunk_encoded + [tokenizer.eos_token_id] * padding_len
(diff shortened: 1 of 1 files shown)


bsmetadata/metadata_utils.py
```diff
@@ -124,16 +124,27 @@ def is_metadata(idx: int) -> bool:
...
-
-            linearized_examples["input_ids"].append(input_ids)
-            linearized_examples["attention_mask"].append(attention_mask)
-            linearized_examples["metadata_mask"].append(metadata_mask)
+            if without_metadata:
+                total_len = len(text_chunk_encoded)
+                padding_len = cfg.max_seq_len - len(text_chunk_encoded)
+                attention_mask = [1] * total_len + [0] * padding_len
+                text_chunk_encoded = text_chunk_encoded + [tokenizer.eos_token_id] * padding_len
+                input_ids = text_chunk_encoded + [tokenizer.eos_token_id] * padding_len
```
cccntu, line 132: `[tokenizer.eos_token_id] * padding_len` is added twice.
Is this a copy-paste mistake?

jordiclive: you are right! Good spot.

PR #189, opened 2023-04-28
Configurable html sample rate, and without metada same context option.

jordiclive:
- Adding html overall sample rate.
- Adding option to run without metadata (with the same context as with metadata)


cccntu (approved):
LGTM

tianjianjiang:
Since @cccntu has reviewed it and the code has been used without functional issues, I'm taking the liberty of merging this PR now.

PR #191, opened 2023-05-14
Update tokenizer in evaluation

manandey:
-> Updated the tokenizer (hardcoded as of now, and would need to change the repo and subfolder if we change it while training again).
-> Commented the code where we were setting the logits of entity related special tokens to -inf as of now until the issue of 'nan's' is fixed.


bsmetadata/evaluation.py
```diff
@@ -348,8 +348,13 @@ def create_metadata_prompt(example: Dict[str, Any], cfg: MetadataConfig) -> str:
...
 
     # Load tokenizer
-    tokenizer = AutoTokenizer.from_pretrained(repo_args.model_name)
-    tokenizer.pad_token = tokenizer.eos_token
+    if args.untrained:
+        tokenizer = AutoTokenizer.from_pretrained(repo_args.model_name)
+        tokenizer.pad_token = tokenizer.eos_token
+    else:
+        tokenizer = AutoTokenizer.from_pretrained(
+            "bs-modeling-metadata/checkpoints_all_04_23", subfolder="tokenizer", use_auth_token=True
```
tianjianjiang, line 356: @manandey I will change this to use the arguments but thank you!

jordiclive: @tianjianjiang Are you looking at adding the HTML special tokens? Then this tokenizer would change.

jordiclive (approved):
lgtm. We can merge this and work on getting to the bottom of the -inf, HTML in the meantime.

bsmetadata/evaluation.py
tianjianjiang: @jordiclive Yes, it's one of the reasons and yet using arguments are probably just more general but of course it doesn't matter for the current version anyway. :)

